PR #13, opened 2017-09-24
Setup basic google login

kevinchan159:
Setup basic GoogleSignIn methods


.gitignore
```diff
@@ -64,6 +64,7 @@ fastlane/Preview.html
 fastlane/screenshots
 fastlane/test_output
 
-Secrets/
+Secrets/GoogleService-Info.plist
```
dantheli: Keep it `Secrets` so that everything inside will be ignored

PR #20, opened 2017-09-26
Added models for User, Course, Session

kevinchan159:
Setup models for User, Course, and Session.


Clicker/Models/Course.swift
```diff
@@ -0,0 +1,49 @@
...
+
+import UIKit
+
+enum SubjectType {
+    // Will add rest later if we plan
+    // on keeping this
+    case CS, INFO, MATH
+}
+
+class Course: NSObject {
```
dantheli: Good work, but what's reasoning for making them NSObject?

kevinchan159: Oh it was because for the user I wanted to keep track of a user's courses in a Set so conforming to NSObject would make it Hashable. For some reason I assumed we aren't displaying the courses of a user. If we actually want to display the courses a user has in like a TableView, though, storing the courses as an array would probably make more sense.

dantheli: I think it would be fine to put them in arrays. When we display them in table views we'll probably copy the arrays inside view controllers

kevinchan159: Alright, I'll change it then recommit

Clicker/Models/Course.swift
```diff
@@ -0,0 +1,49 @@
...
+//  Course.swift
+//  Clicker
+//
+//  Created by Kevin Chan on 9/26/17.
+//  Copyright © 2017 CornellAppDev. All rights reserved.
+//
+
+import UIKit
+
+enum SubjectType {
```
dantheli: Sadly Subject needs to be a String instead
I don't see a lot of benefit in making every possible subject an enum case.

PR #23, opened 2017-10-05
Keivan/livesessions

keivanshahida:
Set up basic design for dashboard and navigation.


Clicker/Views/Cells/LiveSessionTableViewCell.swift
```diff
@@ -0,0 +1,46 @@
...
+        sessionLabel.font = UIFont.systemFont(ofSize: 16, weight: UIFont.Weight.medium)
+
+        self.contentView.addSubview(sessionLabel!)
+        setConstraints()
+    }
+    func setConstraints() {
+        sessionLabel.snp.makeConstraints { (make) -> Void in
+            make.centerY.equalTo(self.contentView)
+            make.left.equalTo(self.contentView).offset(56.5)
+        }
```
dantheli: Because session label is a subview of content view, you can simply call
`make.centerY.equalToSuperview()` etc.

Clicker/Views/Cells/LiveSessionTableViewCell.swift
```diff
@@ -0,0 +1,46 @@
...
+    override open func layoutSubviews() {
+        super.layoutSubviews()
+        
+        sessionLabel = UILabel(frame: .zero)
+        sessionLabel.clipsToBounds = false
+        sessionLabel.textColor = .black
+        sessionLabel.text = "ASTRO 1101 - Lecture 8" //TEMP
+        sessionLabel.font = UIFont.systemFont(ofSize: 16, weight: UIFont.Weight.medium)
+
+        self.contentView.addSubview(sessionLabel!)
```
dantheli: v minor but a little cleaner to do
`weight: .medium`
and remove self in `self.contentView`

Clicker/Views/Cells/LiveSessionTableViewCell.swift
```diff
@@ -0,0 +1,46 @@
...
+//  Created by Keivan Shahida on 10/1/17.
+//  Copyright © 2017 CornellAppDev. All rights reserved.
+//
+
+import UIKit
+import SnapKit
+
+class LiveSessionTableViewCell: UITableViewCell {
+    
+    var sessionLabel: UILabel!
```
dantheli: if we initialize sessionLabel here we don't have to use force unwrapping later

Clicker/Views/Cells/PastSessionTableViewCell.swift
```diff
@@ -0,0 +1,73 @@
...
+            make.left.equalTo(self.contentView).offset(18)
+            make.right.equalTo(self.contentView).offset(18)
+            make.top.equalTo(self.contentView).offset(18)
+        }
+        
+        dateLabel.snp.makeConstraints { (make) -> Void in
+            make.left.equalTo(self.contentView).offset(18)
+            make.right.equalTo(self.contentView).offset(18)
+            make.top.equalTo(self.contentView).offset(44.5)
+        }
```
dantheli: same comments from before apply here

Clicker/Views/Headers/PastSessionHeader.swift
```diff
@@ -0,0 +1,56 @@
...
+    }
+    
+    // MARK: - CONSTRAINTS
+    func setConstraints(){
+        sectionHeaderLabel.snp.makeConstraints { (make) -> Void in
+            make.left.equalTo(self.contentView).offset(18)
+            make.right.equalTo(self.contentView).offset(18)
+            make.top.equalTo(self.contentView).offset(27)
+        }
+    }
```
dantheli: same comments

Clicker/Supporting/Constants.swift
```diff
@@ -0,0 +1,24 @@
...
+struct Constants {
+    struct Screen {
+        static let width = UIApplication.shared.keyWindow?.frame.width
+        static let height = UIApplication.shared.keyWindow?.frame.height
+    }
+    
+    struct Headers {
+        struct Height {
+            static let liveSession = 44.5
+            static let pastSession = 53.5
```
dantheli: Dubious about having constants defined here but should be fine for now
Regardless, these should have type `: CGFloat` instead of double

PR #65, opened 2018-03-04
some more fixes

kevinchan159:
- fix code generation so that it only does it upon Create Poll
- Show pending screen when question end message is emitted
- Dismiss VC when session ends
- detecting live session bug fixed


dantheli:
for some reason as user, the socket doesn't seem to be disconnecting properly when the navigationcontroller pops

kevinchan159:
i made some breakpoints and it seemed to be disconnecting on my end. where do you see it staying connected?

PR #116, opened 2018-04-25
Jack/one tab

jschluger:
Switching to one tab design, as we discussed at the meeting today.

### DIFF
--- a/Clicker/View Controllers/LoginViewController.swift
+++ b/Clicker/View Controllers/LoginViewController.swift
@@ -13,7 +13,6 @@ import SnapKit
 class LoginViewController: UIViewController, GIDSignInUIDelegate {
     
     var signInButton: GIDSignInButton!
-    var continueButton: UIButton!
     
     // MARK: - INITIALIZATION
     override func viewDidLoad() {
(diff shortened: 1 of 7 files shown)


Clicker/View Controllers/LoginViewController.swift
```diff
@@ -25,16 +26,31 @@ class LoginViewController: UIViewController, GIDSignInUIDelegate {
         signInButton.colorScheme = .light
         view.addSubview(signInButton)
         
+        continueButton = UIButton()
+        continueButton.setTitle("Continue", for: .normal)
+        continueButton.addTarget(self, action: #selector(continueAction), for: .touchUpInside)
+        view.addSubview(continueButton)
+
```
kevinchan159: Is this the button to fix the moving to home screen after login issue? If so, I think I know how to fix it. We don't need to add an extra button. Theres a didSignIn delegate method that we can insert. Once they log in, we can pop the LoginVC in there.

kevinchan159: Take a look at PollsCell for an example. I think I have that method in there

kevinchan159:
The rest looks good! Feel free to merge after looking at my comment above about the continueButton. If you want, you can try adding the didSignIn method to LoginVC. If you can't figure it out, you can just remove the continueButton code and I can add it sometime this week.

PR #140, opened 2018-08-26
Kevin/extension refactoring

kevinchan159:
Added extension files for view controllers to hold collection view methods

Link to Issue: https://github.com/cuappdev/clicker-ios/issues/137


jschluger:
Code seems good, but what did you change to make the Travis CI builds fail?

The error is "xcodebuild: error: Failed to build workspace Clicker with scheme ClickerTests.
	Reason: A build only device cannot be used to run this target."

kevinchan159:
I don't think I changed anything major? hmm. The odd thing is that I don't think we have any ClickerTests.

PR #146, opened 2018-08-27
Jack/constants refactoring

jschluger:


Clicker/Supporting/Constants.swift
```diff
@@ -9,6 +9,17 @@
 import Foundation
 import UIKit
 
+let MULTIPLE_CHOICE = "MULTIPLE_CHOICE"
```
AAAstorga: can this be `let multipleChoice`?

Clicker/Supporting/Constants.swift
```diff
@@ -9,6 +9,17 @@
...
+let FREE_RESPONSE = "FREE_RESPONSE"
```
AAAstorga: same ^

Clicker/Supporting/Constants.swift
```diff
@@ -9,6 +9,17 @@
...
+let addMoreOptionCellID = "addMoreOptionCellID"
+let createMCOptionCellID = "createMCOptionCellID"
+
+let server_poll_end = "server/poll/end"
+let server_poll_results = "server/poll/results"
+let server_poll_tally = "server/poll/tally"
+let server_poll_upvote = "server/poll/upvote"
+
 enum Keys: String {
     case apiURL = "api-url"
```
AAAstorga: lets follow camel case here and do `case apiUrl`

Clicker/Supporting/Constants.swift
```diff
@@ -9,6 +9,17 @@
...
     case apiDevURL = "api-dev-url"
```
AAAstorga: same ^

Clicker/Supporting/Constants.swift
```diff
@@ -9,6 +9,17 @@
...
+let FREE_RESPONSE = "FREE_RESPONSE"
+let significantEventsIdentifier = "significantEvents"
+let addMoreOptionCellID = "addMoreOptionCellID"
+let createMCOptionCellID = "createMCOptionCellID"
+
+let server_poll_end = "server/poll/end"
+let server_poll_results = "server/poll/results"
+let server_poll_tally = "server/poll/tally"
+let server_poll_upvote = "server/poll/upvote"
+
```
AAAstorga: why do the structs have private empty inits?

Clicker/View Controllers/DraftsViewController.swift
```diff
@@ -14,6 +14,9 @@ protocol FillsDraftDelegate {
 
 class DraftsViewController: UIViewController, UICollectionViewDataSource, UICollectionViewDelegate, UICollectionViewDelegateFlowLayout {
     
+    // MARK: Constants
+    let draftCellID = "draftCellID"
```
AAAstorga: if you want (optional), you should move this to a Constants.ReuseIdentifiers struct.

Clicker/View Controllers/PollBuilderViewController.swift
```diff
@@ -246,17 +246,17 @@ class PollBuilderViewController: UIViewController, QuestionDelegate, PollBuilder
...
         
         // MULTIPLE CHOICE
-        if (questionType == "MULTIPLE_CHOICE") {
+        if (questionType == MULTIPLE_CHOICE) {
 
             let question = mcPollBuilder.questionTextField.text
             let options = mcPollBuilder.optionsDict.keys.sorted().map { mcPollBuilder.optionsDict[$0]! }
             
-            startPollDelegate.startPoll(text: question!, type: "MULTIPLE_CHOICE", options: options, isShared: false)
+            startPollDelegate.startPoll(text: question!, type: MULTIPLE_CHOICE, options: options, isShared: false)
```
AAAstorga: don't force unwrap these. At the minimum do `let question = mcPollBuilder.questionTextField.text ?? ""`, but think about the best option here

Clicker/View Controllers/PollBuilderViewController.swift
```diff
@@ -246,17 +246,17 @@ class PollBuilderViewController: UIViewController, QuestionDelegate, PollBuilder
...
         } else { // FREE RESPONSE
             
             let question = frPollBuilder.questionTextField.text
             let isShared = frPollBuilder.dropDown.shareResponses
-            startPollDelegate.startPoll(text: question!, type: "FREE_RESPONSE", options: [], isShared: isShared)
+            startPollDelegate.startPoll(text: question!, type: FREE_RESPONSE, options: [], isShared: isShared)
```
AAAstorga: ^ same as above

Clicker/Supporting/Constants.swift
```diff
@@ -9,6 +9,17 @@
 import Foundation
 import UIKit
 
+let MULTIPLE_CHOICE = "MULTIPLE_CHOICE"
+let FREE_RESPONSE = "FREE_RESPONSE"
+let significantEventsIdentifier = "significantEvents"
+let addMoreOptionCellID = "addMoreOptionCellID"
+let createMCOptionCellID = "createMCOptionCellID"
+
+let server_poll_end = "server/poll/end"
```
AAAstorga: these should probably follow camel case not snake case

Clicker/Supporting/Constants.swift
kevinchan159: Its to prevent anyone from initializing them I believe

Clicker/Views/OptionsView.swift
```diff
@@ -15,6 +15,9 @@ protocol SliderBarDelegate {
 
 class OptionsView: UIView, UICollectionViewDataSource, UICollectionViewDelegate, UICollectionViewDelegateFlowLayout {
     
+    //MARK: Constants
+    let questionOptionCellId = "questionOptionCellId"
```
kevinchan159: We should probably be consistent on where we're placing constants. 
I think one way we could do it is place all the cell identifiers into the Constants file under the ReuseIdentifiers struct but keep constants like CGFloats as instance variables of the class since they're a bit more specific to the class.

kevinchan159:
Overall, LGTM. Just address @AAAstorga comments and take a look at mine before merging. Nice job!!!

jschluger:
I improved the way we store constants taking into consideration this feedback. 

Now there is a `QuestionType` which includes a `.description` property to easily convert the question type to a string, as we do many places in the app. 

There is now an Identifiers struct with static variables holding the values of identifiers used in the project.

Finally, there is now a Routes struct which contains static variables holding string values of the route names used to reach the backend. 

@AAAstorga @kevinchan159

Clicker/Views/HomeScreen/Polls/PollsCell.swift
```diff
@@ -16,10 +16,14 @@ protocol EditSessionDelegate {
 
 class PollsCell: UICollectionViewCell, UITableViewDelegate, UITableViewDataSource, EditPollDelegate, GIDSignInDelegate {
     
+    // MARK: Constants
+    let pollPreviewIdentifier = "pollPreviewCellID"
+    let adminIdentifier = "admin"
```
kevinchan159: Can you move these identifiers to the Identifiers struct as well?

kevinchan159:
Looks awesome! Just one more minor change, but should be good to merge after. Great job :)

jschluger:
I made these changes and migrated some more identifiers that I previously missed.

PR #147, opened 2018-08-27
Converted home collection view to IGListKit

kevinchan159:
- Converted home collection view to use IGListKit.
ListDiffable Model: PollTypeModel
ListSectionController: PollTypeSectionController

- Renamed `View Controllers` directory to `ViewControllers`
- Created more extensions for `PollsViewController` to separate delegate methods


Clicker/SectionControllers/PollTypeSectionController.swift
```diff
@@ -0,0 +1,54 @@
+//
+//  PollTypeSectionController.swift
+//  Clicker
+//
+//  Created by Kevin Chan on 8/27/18.
+//  Copyright © 2018 CornellAppDev. All rights reserved.
+//
+
+import Foundation
+import IGListKit
```
AAAstorga: I love IGListKit <3

Clicker/Models/ListDiffableModels/PollTypeModel.swift
```diff
@@ -0,0 +1,37 @@
...
+    }
+    
+    func diffIdentifier() -> NSObjectProtocol {
+        return identifier as NSString
+    }
+    
+    func isEqual(toDiffableObject object: ListDiffable?) -> Bool {
+        if (self === object) { return true }
+        guard let object = object as? PollTypeModel else { return false }
+        return pollType == object.pollType
```
AAAstorga: shouldn't you be comparing the identifiers to determine if two PollTypeModels are equal?

Clicker/SectionControllers/PollTypeSectionController.swift
```diff
@@ -0,0 +1,54 @@
...
+    var delegate: PollTypeSectionControllerDelegate!
+    
+    init(delegate: PollTypeSectionControllerDelegate) {
+        self.delegate = delegate
+    }
+    
+    // MARK: - ListSectionController overrides
+    override func sizeForItem(at index: Int) -> CGSize {
+        // This is the size of the whole collection view
+        return (collectionContext?.containerSize)!
```
AAAstorga: can we safely unwrap this. Like `guard let context = collectionContext else { return .zero }`

Clicker/ViewControllers/PollsViewController+Extension.swift
```diff
@@ -0,0 +1,76 @@
...
+        return nil
+    }
+    
+    // MARK: - PollTypeSectionControllerDelegate
+    var editSessionDelegate: EditSessionDelegate {
+        return self
+    }
+    
+    func sectionControllerWillDisplayPollType(sectionController: PollTypeSectionController, pollType: PollType) {
+        let item: Int = pollType == .created ? 0 : 1
```
AAAstorga: use type inference here

Clicker/Models/ListDiffableModels/PollTypeModel.swift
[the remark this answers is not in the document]
kevinchan159: The way I have it defined, the identifiers for any two instances are going to be different. If the pollTypes are equal (i.e. they are both `.created`) then we should consider them to be equal

Clicker/SectionControllers/PollTypeSectionController.swift
kevinchan159: yep will do

Clicker/SectionControllers/PollTypeSectionController.swift
```diff
@@ -0,0 +1,54 @@
...
+    override func didUpdate(to object: Any) {
+        pollTypeModel = object as? PollTypeModel
+    }
+    
+    // MARK: - ListDisplayDelegate
+    func listAdapter(_ listAdapter: ListAdapter, willDisplay sectionController: ListSectionController) {
+        delegate.sectionControllerWillDisplayPollType(sectionController: self, pollType: pollTypeModel.pollType)
+    }
+    
+    func listAdapter(_ listAdapter: ListAdapter, didEndDisplaying sectionController: ListSectionController) {}
```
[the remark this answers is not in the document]
kevinchan159: I don't think so but the delegate requires us to implement these 3 methods as well

Clicker/SectionControllers/PollTypeSectionController.swift
kevinchan159: <3

PR #148, opened 2018-08-30
Code cleanup before migrating CardController to IGListKit

kevinchan159:
Some code style cleanup before redoing CardController with IGListKit.


kevinchan159:
Alt Text

PR #149, opened 2018-09-02
Migrating Vertical CollectionView to IGListKit

kevinchan159:
First off, sorry for the big PR. Didn't want to submit a PR before I finished migrating the whole vertical collection view. PR is mainly big because of the new files.
In this PR:
- Migrated vertical collection view to IGListKit
- Improved Poll.swift model
- Cleaned up UIColor+Shared.swift
- Added new struct for LayoutConstants to Constants.swift

Screenshot:
screen shot 2018-09-01 at 8 53 01 pm


Clicker/AppDelegate.swift
```diff
@@ -23,7 +23,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, GIDSignInDelegate {
     func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplicationLaunchOptionsKey: Any]?) -> Bool {
         
         window = UIWindow(frame: UIScreen.main.bounds)
-        window?.tintColor = .clickerGreen
+        window?.tintColor = .clickerGreen0
```
AAAstorga: can we get better names then `clickerGreen0`. What does the # mean?

Clicker/Extensions/UIColor+Shared.swift
```diff
@@ -9,31 +9,28 @@ import UIKit
...
+    @nonobjc static let clickerGrey3 = UIColor(red: 122/255, green: 129/255, blue: 139/255, alpha: 1.0)
+    @nonobjc static let clickerGrey4 = UIColor(red: 247/255, green: 249/255, blue: 250/255, alpha: 1.0)
+    @nonobjc static let clickerGrey5 = UIColor(red: 225/255, green: 225/255, blue: 225/255, alpha: 1.0)
+    @nonobjc static let clickerGrey6 = UIColor(red: 240/255, green: 240/255, blue: 240/255, alpha: 1.0)
+    @nonobjc static let clickerGrey7 = UIColor(red: 246/255, green: 246/255, blue: 248/255, alpha: 1.0)
+    @nonobjc static let clickerGrey8 = UIColor(red: 249/255, green: 249/255, blue: 249/255, alpha: 1.0)
+    @nonobjc static let clickerGrey9 = UIColor(red: 32/255, green: 32/255, blue: 32/255, alpha: 0.6)
+    @nonobjc static let clickerGrey10 = UIColor(red: 61/255, green: 61/255, blue: 61/255, alpha: 1.0)
+    @nonobjc static let clickerBlack0 = UIColor(red: 31/255, green: 44/255, blue: 56/255, alpha: 1.0)
+    @nonobjc static let clickerBlack1 = UIColor(red: 32/255, green: 32/255, blue: 32/255, alpha: 1.0)
```
AAAstorga: not sure how I feel about the color names lol (don't hate me). Lets talk about what you think. Why do you think this is a better approach?

Clicker/Models/ListDiffableModels/Poll.swift
```diff
@@ -0,0 +1,115 @@
...
+}
+
+class Poll {
+    
+    var id: Int?
+    var text: String
+    var questionType: QuestionType
+    var options: [String]
+    var results: [String:Any]
+    // results format:
```
AAAstorga: move state to be above results, so the results comment is at the end of the block

Clicker/Models/ListDiffableModels/Poll.swift
```diff
@@ -0,0 +1,115 @@
...
+    let countKey = "count"
+    let optionsKey = "options"
+    let typeKey = "type"
+    
+    // MARK: SORTED BY DATE POLL INITIALIZER
+    init(id: Int, text: String, results: [String:Any], type: QuestionType, state: PollState) {
+        self.id = id
+        self.text = text
+        self.results = results
+        self.options = results.map { (key, _) in key }
```
AAAstorga: can't you do `results.keys`

Clicker/Models/ListDiffableModels/Poll.swift
```diff
@@ -0,0 +1,115 @@
...
+        self.state = state
+    }
+    
+    // MARK: SEND START POLL INITIALIZER
+    init(text: String, options: [String], type: QuestionType, state: PollState) {
+        self.text = text
+        self.options = options
+        self.results = [:]
+        self.questionType = type
+        for (index, option) in options.enumerated() {
```
AAAstorga: thoughts on this vs `options.enumerated().forEach {...`

Clicker/Models/ListDiffableModels/Poll.swift
```diff
@@ -0,0 +1,115 @@
...
+            let mcOption = intToMCOption(index)
+            results[mcOption] = [textKey: option, countKey: 0]
+        }
+        self.state = state
+    }
+    
+    // MARK: RECEIVE START POLL INITIALIZER
+    init(json: [String:Any]){
+        self.id = json[idKey] as? Int
+        self.text = json[textKey] as! String
```
AAAstorga: don't force this, unwrap it properly

Clicker/Models/ListDiffableModels/Poll.swift
```diff
@@ -0,0 +1,115 @@
...
+        self.state = .live
+    }
+    
+    // Returns array representation of results
+    // Ex) [('Blah', 3), ('Jupiter', 2)...]
+    func getFRResultsArray() -> [(String, Int)] {
+        var resultsArr: [(String, Int)] = []
+        results.forEach { (key, val) in
+            if let choiceJSON = val as? [String:Any] {
+                resultsArr.append((key, (choiceJSON[countKey] as! Int)))
```
AAAstorga: no force unwrapping

Clicker/Models/ListDiffableModels/Poll.swift
```diff
@@ -0,0 +1,115 @@
...
+        }
+        return resultsArr
+    }
+    
+    func getTotalResults() -> Float {
+        return results.reduce(0) { (res, arg1) -> Float in
+            let (_, value) = arg1
+            if let choiceJSON = value as? [String:Any], let numSelected = choiceJSON[countKey] as? Float {
+                return res + numSelected
+            } else {
```
AAAstorga: don't need else, since you return in the if let

Clicker/SectionControllers/PollDateSectionController.swift
```diff
@@ -0,0 +1,92 @@
...
+    override func cellForItem(at index: Int) -> UICollectionViewCell {
+        var cell: UICollectionViewCell
+        let itemType = itemTypes[index]
+        switch (itemType) {
+        case .date:
+            let dateCell = collectionContext?.dequeueReusableCell(of: DateCell.self, for: self, at: index) as! DateCell
+            dateCell.configure(with: pollDateModel.date)
+            dateCell.setNeedsUpdateConstraints()
+            cell = dateCell
+            break
```
AAAstorga: do you need break here? I forget if its actually needed

Clicker/SectionControllers/PollDateSectionController.swift
```diff
@@ -0,0 +1,92 @@
...
+        case .card:
+            let cardCellHeight = calculateCardCellHeight(poll: pollDateModel.poll, userRole: delegate.role)
+            return CGSize(width: containerSize.width, height: cardCellHeight)
+        }
+    }
+    
+    override func cellForItem(at index: Int) -> UICollectionViewCell {
+        var cell: UICollectionViewCell
+        let itemType = itemTypes[index]
+        switch (itemType) {
```
AAAstorga: don't need parentheses

Clicker/ViewControllers/CardController+Extension.swift
```diff
@@ -9,6 +9,47 @@
...
+        switch state {
+        case .horizontal:
+            if (currentIndex > -1) {
+                return pollsDateArray[currentIndex].polls
+            } else {
+                return [EmptyStateModel(userRole: userRole)]
+            }
+        default:
+            return pollsDateArray.map({ pollsDateModel -> PollDateModel in
+                let latestPoll = pollsDateModel.polls[pollsDateModel.polls.count - 1]
```
AAAstorga: use pollsDateModel.polls.last and unwrAp

Clicker/ViewControllers/CardController+Extension.swift
```diff
@@ -9,6 +9,47 @@
...
+            })
+        }
+    }
+    
+    func listAdapter(_ listAdapter: ListAdapter, sectionControllerFor object: Any) -> ListSectionController {
+        if object is Poll {
+            return PollSectionController()
+        } else if object is PollDateModel {
+            return PollDateSectionController(delegate: self)
+        } else {
```
AAAstorga: no need for else

Clicker/ViewControllers/CardController.swift
```diff
@@ -326,7 +382,7 @@ class CardController: UIViewController {
     
     @objc func detectedPinchAction(_ sender: UIPinchGestureRecognizer) {
         let isPinchOut: Bool = (sender.scale > 1)
-        if (isPinchOut && verticalCollectionView != nil && !verticalCollectionView.isDescendant(of: self.view)) {
+        if (isPinchOut) {
```
AAAstorga: no (...)

Clicker/Extensions/UIColor+Shared.swift
kevinchan159: The reason I changed this was because before what we would have are names like clickerLightGrey, clickerMediumGrey, clickerDarkGrey, etc. The two problems I had with this is: - we would then have 2 other grey colors, with no real good way left to name them
- the idea behind naming these colors `clickerMediumGrey` is so that we know which grey to use when I look at the design but in reality, I always just go back to this file and compare the RGB values b/c I don't remember which type of grey it was.
So I thought that just naming it by number would solve the issue of coming up with a descriptive name.

Clicker/Models/ListDiffableModels/Poll.swift
kevinchan159: I like the `forEach` :)

Clicker/SectionControllers/PollDateSectionController.swift
kevinchan159: oh you don't. force of habit from last semester haha. ill remove it

Clicker/Extensions/UIColor+Shared.swift
kevinchan159: LMK what you think about ^. curious if theres a better way

jschluger: I agree that this color naming convention isn't any worse than the previous convention including `clickerMediumGrey`, because during development I also always check the RGB values on the design to make sure to use the correct color. Additionally this new convention allows an arbitrary number of colors to be defined with consistent names, a positive feature.

Clicker/Quarks/PollQuark.swift
```diff
@@ -176,7 +180,8 @@ struct UpdatePoll: ClickerQuark {
                 throw NeutronError.badResponseData
             }
             let questionType: QuestionType = (type == Identifiers.multipleChoiceIdentifier) ? .multipleChoice : .freeResponse
-            return Poll(id: id, text: text, results: results, type: questionType, isShared: shared)
+            let state: PollState = shared ? .shared : .ended
```
jschluger: I appreciate the use of enums in this file instead of passing the boolean through to the Poll initializer. Nice!

PR #158, opened 2018-09-15
Jack/settings screen

jschluger:
Resolve issue #154, building the Settings screen.
This includes:
 - implementing Designs for the settings screen
 - Implementing signing out with the Google Sign In library

Note: this PR implements signing out from the perspective of the models, but does not do anything when the user signs out to show them that they are signed out (like returning to the sign in screen), and instead lets them keep using the app. However, I think we should have one issue just for solidifying the entire sign in and out flow, which would solve this problem. 

image


Clicker/Models/ListDiffableModels/SettingsDataModel.swift
```diff
@@ -0,0 +1,43 @@
...
+import IGListKit
+
+enum SettingsDataState {
+    case link // just a link to something
+    case info // a title and some information
+}
+
+class SettingsDataModel {
+    
+    let identifier = UUID().uuidString
```
kevinchan159: can you put `identifier` below the `vars`?

Clicker/SectionControllers/SettingsSectionController.swift
```diff
@@ -0,0 +1,49 @@
...
+        guard let containerSize = collectionContext?.containerSize else {
+            return .zero
+        }
+        switch settingsDataModel.state {
+        case .info:
+            return CGSize(width: containerSize.width, height: settingsDataModel.title == "About" ? 110 : 87)
+        case .link:
+            return CGSize(width: containerSize.width, height: 39.5)
+        }
+    }
```
kevinchan159: make each of the constants in this function a class variable, i.e.:
`let cellHeight = 110`

Clicker/ViewControllers/PollsViewController.swift
```diff
@@ -120,7 +126,14 @@ class PollsViewController: UIViewController {
...
             make.height.equalTo(19)
-            make.right.equalToSuperview().offset(-15)
+            make.right.equalToSuperview().inset(15)
+        }
+        
+        settingsButton.snp.makeConstraints { make in
+            make.height.equalTo(30)
+            make.width.equalTo(30)
+            make.centerY.equalTo(newPollButton.snp.centerY)
+            make.left.equalToSuperview().offset(15)
```
kevinchan159: make these numbers class `let` variables

Clicker/ViewControllers/SettingsViewController.swift
```diff
@@ -0,0 +1,152 @@
...
+        setupViews()
+        setupConstraints()
+    }
+    
+    func loadData() {
+        let s1 = SettingsDataModel(state: .info, title: "Account", description: User.currentUser?.email)
+        let s2 = SettingsDataModel(state: .info, title: "About", description:
+            "Pollo is made by Cornell AppDev, an engineering project team at Cornell University.")
+        let s3 = SettingsDataModel(state: .link, title: "More Apps", description: "https://www.cornellappdev.com/")
+        let s4 = SettingsDataModel(state: .link, title: "Visit Our Website")
```
kevinchan159: make these strings class `let` variables. also can we name it to something like `settingsModel4` instead of `s4`, just for readibility

Clicker/Views/Settings/SettingsInfoCell.swift
```diff
@@ -0,0 +1,80 @@
...
+        descriptionView.snp.makeConstraints { make in
+            make.top.equalTo(titleView.snp.bottom).offset(6)
+            make.left.equalTo(titleView.snp.left)
+            make.bottom.equalToSuperview().inset(20.5)
+            make.right.equalTo(titleView.snp.right)
+        }
+        
+        lineView.snp.makeConstraints { make in
+            make.height.equalTo(1)
+            make.left.equalToSuperview().offset(18)
```
kevinchan159: make constants class `let` variables

PR #166, opened 2018-09-18
Fixed keyboard bug and modified design of question type drop-down

matthewcoufal:
img_1751
I used Presentr to make this drop down from the top of the screen. The only problem is that upon dismissal the view goes down instead of back up (but the view itself still disappears). Let me know if there's a workaround that could fix this or if it's not that big of a deal.
Also, the keyboard bug was related to the way we were modifying the position of buttonsView. Before we directly modified its frame, but now we update its constraints using SnapKit. The view was disappearing before because it somehow got updated to its original constraints. But now this is not a problem.


Clicker/ViewControllers/CardController.swift
```diff
@@ -290,7 +290,9 @@ class CardController: UIViewController {
         presenter.dismissOnSwipeDirection = .bottom
         
         let nc = UINavigationController(rootViewController: pollBuilderVC)
-        customPresentViewController(presenter, viewController: nc, animated: true, completion: nil)
+        nc.isNavigationBarHidden = true
```
kevinchan159: I know that you didn't name the variable but can you rename `nc` to something like `pollBuilderNavigationController` or something like that

Clicker/ViewControllers/EditQuestionTypeViewController.swift
```diff
@@ -0,0 +1,124 @@
...
+import UIKit
+
+protocol EditQuestionTypeDelegate {
+    func didPick(questionType: QuestionType)
+}
+
+class EditQuestionTypeViewController: UIViewController {
+
+    var delegate: EditQuestionTypeDelegate?
+    var selectedQuestionType: QuestionType!
```
kevinchan159: put these in between `View vars` and `Constants` and add a mark for it called`// MARK: - Data vars`

Clicker/ViewControllers/EditQuestionTypeViewController.swift
```diff
@@ -0,0 +1,124 @@
...
+        super.viewDidLoad()
+        
+        view.backgroundColor = .white
+
+        multipleChoiceButton = UIButton()
+        multipleChoiceButton.addTarget(self, action: #selector(didPickMultipleChoice), for: .touchUpInside)
+        view.addSubview(multipleChoiceButton)
+        
+        multipleChoiceLabel = UILabel()
+        multipleChoiceLabel.text = "Multiple Choice"
```
kevinchan159: Add `let multipleChoiceLabelText = "MultipleChoice"` as a class variable and use that instead

Clicker/ViewControllers/EditQuestionTypeViewController.swift
```diff
@@ -0,0 +1,124 @@
...
+        multipleChoiceLabel.textColor = selectedQuestionType == .multipleChoice ? .aquaMarine : .clickerBlack0
+        multipleChoiceLabel.font = selectedQuestionType == .multipleChoice ? ._16SemiboldFont : ._16RegularFont
+        view.addSubview(multipleChoiceLabel)
+        
+        freeResponseButton = UIButton()
+        freeResponseButton.addTarget(self, action: #selector(didPickFreeResponse), for: .touchUpInside)
+        view.addSubview(freeResponseButton)
+        
+        freeResponseLabel = UILabel()
+        freeResponseLabel.text = "Free Response"
```
kevinchan159: same here

Clicker/ViewControllers/EditQuestionTypeViewController.swift
```diff
@@ -0,0 +1,124 @@
...
+        selectedDot.clipsToBounds = true
+        selectedDot.layer.cornerRadius = dotDiameter / 2
+        view.addSubview(selectedDot)
+        
+        layoutSubviews()
+    }
+    
+    func layoutSubviews() {
+        
+        multipleChoiceButton.snp.makeConstraints { make in
```
kevinchan159: Instead of putting these constraints in `layoutSubviews`, can you just make a function called `setupConstraints` and put them in there? or do you need to have it in `layoutSubviews`?

Clicker/ViewControllers/EditQuestionTypeViewController.swift
```diff
@@ -0,0 +1,124 @@
...
+//  Clicker
+//
+//  Created by Matthew Coufal on 9/18/18.
+//  Copyright © 2018 CornellAppDev. All rights reserved.
+//
+
+import UIKit
+
+protocol EditQuestionTypeDelegate {
+    func didPick(questionType: QuestionType)
```
kevinchan159: Can you rename this function to be `editQuestionTypeViewControllerDidPick(...)`. This helps us to easily tell which classes these delegate methods belong to

Clicker/ViewControllers/PollBuilderViewController.swift
```diff
@@ -315,8 +316,31 @@ class PollBuilderViewController: UIViewController, QuestionDelegate, PollBuilder
...
+        updateQuestionTypeButton()
+        switch questionType {
+        case .multipleChoice:
+            mcPollBuilder.isHidden = false
+            frPollBuilder.isHidden = true
+            break
+        case .freeResponse:
+            mcPollBuilder.isHidden = true
+            frPollBuilder.isHidden = false
+        }
```
kevinchan159: You can reduce this down to:
`mcPollBuilder.isHidden = questionType == .freeResponse`
`frPollBuilder.isHidden = questionType == . multipleChoice`

PR #182, opened 2018-09-23
Added empty state to draftsVC

matthewcoufal:
simulator screen shot - iphone 8 plus - 2018-09-22 at 21 47 31


Clicker/ViewControllers/DraftsViewController+Extension.swift
```diff
@@ -18,18 +18,30 @@ extension DraftsViewController: DraftCellDelegate {
 extension DraftsViewController: ListAdapterDataSource {
     
     func objects(for listAdapter: ListAdapter) -> [ListDiffable] {
-        return drafts
+        if drafts.count > 0 {
+            return drafts
+        }
+        return [EmptyStateModel(type: .draftsViewController(delegate: self))]
```
kevinchan159: shorten to `return drafts.count > 0 ? drafts : [EmptyStateModel(type: .draftsViewController(delegate: self))]`

PR #185, opened 2018-09-25
More bugfixes

kevinchan159:
This PR fixes:
- When I try to delete a group that I’ve created … It says “Deleting will permanently close the group for all participants and all poll data will be lost” … we only want that to be the case for groups that I’ve created … for groups that are joined we should only be leaving them as a member … 
- When I join a group using a code … the code remains in that text field 
- Buttons on the Edit Session are too close to bottom on iPhoneX
- Delete (…) button for poll
- The only two concepts that the user should be aware of are “Groups” and “Polls” @Kevin C make sure we don’t see any other names in the app besides those two

In terms of app structure changes:
- When joining a group → show sessions screen (the dates w/ number of question for each date) → then clicking on a date should show a new screen of the polls for that day
    - Did this by separating CardController -> PollsDateViewController (the sessions screen) and Card Controller

### DIFF
--- a/Clicker/ViewControllers/CardController+Extension.swift
+++ b/Clicker/ViewControllers/CardController+Extension.swift
@@ -164,6 +164,7 @@ extension CardController: SocketDelegate {
     func sessionDisconnected() {}
     
     func receivedUserCount(_ count: Int) {
+        numberOfPeople = count
         peopleButton.setTitle("\(count)", for: .normal)
     }
     
--- a/Clicker/ViewControllers/CardController+Extension.swift
+++ b/Clicker/ViewControllers/CardController+Extension.swift
@@ -178,7 +178,7 @@ extension CardController: SocketDelegate {
     }
     
     func pollEnded(_ poll: Poll, userRole: UserRole) {
-        guard let latestPoll = getLatestPoll() else { return }
+        guard let latestPoll = pollsDateModel.polls.last else { return }
         if userRole == .admin {
             latestPoll.id = poll.id
             updateLatestPoll(with: latestPoll)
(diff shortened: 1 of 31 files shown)


Clicker/ViewControllers/CardController+Extension.swift
```diff
@@ -13,31 +13,12 @@ import UIKit
...
-            if (currentIndex > -1) {
-                return pollsDateArray[currentIndex].polls
-            } else {
-                let type: EmptyStateType = .cardController(userRole: userRole)
-                return [EmptyStateModel(type: type)]
-            }
-        default:
-            return pollsDateArray
-        }
+        return pollsDateModel.polls
```
jschluger: So much cleaner!

Clicker/ViewControllers/CardController.swift
```diff
@@ -53,47 +49,43 @@ class CardController: UIViewController {
...
-    func setupCards() {
+    func setupViews() {
         collectionViewLayout = UICollectionViewFlowLayout()
         collectionViewLayout.minimumInteritemSpacing = 0
         collectionViewLayout.minimumLineSpacing = 0
         collectionViewLayout.scrollDirection = .horizontal
         collectionView = UICollectionView(frame: .zero, collectionViewLayout: collectionViewLayout)
-        let collectionViewInset = view.frame.width * 0.05
-        collectionView.contentInset = UIEdgeInsetsMake(0, collectionViewInset, 0, collectionViewInset)
+        let collectionViewXInset = view.frame.width * 0.05
```
jschluger: What does the `X` refer to?

PR #188, opened 2018-09-26
Bug Fixing

jschluger:
In this PR I fix some bugs in the app:
- Fix Timer Bug (#168). 
  - I am interested in your feedback on how I did this, my solution might seem wrong at first: I added the timer stuff to the `Poll` model itself, rather than to the collection view cell displaying that poll. But, I think this does actually make sense conceptually because the timer is associated with the Poll, not with any specific collection view cell which must be reusable for many polls.
- Fixed bug where a user could press the (+) button twice rapidly from the home page to create strange behavior.
- Increased the size of the (+) button on the home screen, and the (...) buttons on Group Previews on the home screen.
- Improved the way the collection view in `CardController` scrolls, so that it is more "sensitive," because previously you had to swipe your finger very far across the screen to get to the next cell, which feels unnatural.

### DIFF
--- a/Clicker/ViewControllers/PollsViewController.swift
+++ b/Clicker/ViewControllers/PollsViewController.swift
@@ -201,6 +201,7 @@ class PollsViewController: UIViewController {
     
     // MARK - Actions
     @objc func newPollAction() {
+        newPollButton.isEnabled = false
         GenerateCode().make()
             .done { code in
                 StartSession(code: code, name: code, isGroup: false).make()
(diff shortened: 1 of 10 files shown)


Clicker/ViewControllers/PollsViewController.swift
```diff
@@ -186,8 +187,8 @@ class PollsViewController: UIViewController {
         
         newPollButton.snp.makeConstraints { make in
             make.top.equalTo(view.safeAreaLayoutGuide.snp.top).offset(15)
-            make.width.equalTo(19)
-            make.height.equalTo(19)
+            make.width.equalTo(29)
+            make.height.equalTo(29)
```
kevinchan159: Can you make these constants?

PR #192, opened 2018-09-29
Join Sessions

kevinchan159:
Update quarks for backend changes and fix poll builder screen


Clicker/Extensions/Utils.swift
```diff
@@ -8,6 +8,7 @@
 
 import Foundation
 import SwiftyJSON
+import UIKit
```
jschluger: Why does this need UIKit now?

PR #194, opened 2018-10-03
mo bugfixes

kevinchan159:
- Change to `Yes, Leave` for members
- Multiple date for members
- Members receiving an extra question 
- Scroll to the end of card controller
- Going into a session with an older poll still shows todays date
- Create new group → start 2 questions. then starting another question creates another date session for the question
- MC options displayed squashed (I think i fixed this, not too sure though b/c I'm not sure how to reproduce 100% of the time)
- 2 questions were created in one session and they were split into two “September 30” sessions with one question in each session
- Home Page: Remove the ability for a user to “edit name” and “delete” on a JOINED poll 

NOTE: Some of these fixes were done assuming some backend changes. Backend will merge their branches into master and then I will merge this branch in after

### DIFF
--- a/Clicker/Quarks/PollQuark.swift
+++ b/Clicker/Quarks/PollQuark.swift
@@ -99,7 +99,7 @@ struct GetSortedPolls: ClickerQuark {
                 }
             }
             let formatter = DateFormatter()
-            formatter.dateFormat = "MMM dd yyyy"
+            formatter.dateFormat = StringConstants.dateFormat
             pollsDateArray.sort { (pollsDateModelA, pollsDateModelB) -> Bool in
                 let dateA = formatter.date(from: pollsDateModelA.date) ?? Date()
                 let dateB = formatter.date(from: pollsDateModelB.date) ?? Date()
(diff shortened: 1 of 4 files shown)


young-k:
Can you squash some of the commits like "Fix merge conflict", plus you spelled "Session" wrong in "LeaveSession".

In general you should try and have fewer commits per PR, I think 3-4 should be a reasonable number

Clicker/Quarks/PollQuark.swift
```diff
@@ -94,12 +94,12 @@ struct GetSortedPolls: ClickerQuark {
...
                         return PollParser.parseItem(json: $0)
                     }
-                    let pollsDateModel = PollsDateModel(date: date, polls: pollsArr)
+                    let pollsDateModel = PollsDateModel(date: date.trimmingCharacters(in: .whitespaces), polls: pollsArr)
                     pollsDateArray.append(pollsDateModel)
                 }
             }
             let formatter = DateFormatter()
-            formatter.dateFormat = "dd/MM/yyyy"
+            formatter.dateFormat = "MMM dd yyyy"
```
jschluger: Make a constant for the string `"MM dd yyyy"`, since we use it in a few different places.

PR #195, opened 2018-10-03
Jack/more bug fixes

jschluger:
In this commit I fix some of the UI bugs. I...
 - Change the caret color across the entire app to Clicker Green.
 - Change the logic of pressing go back from `PollsDataViewController` so that if the session needs to be deleted, this is done after popping back to the home screen, so the user doesn't have to wait for a response from the server to continue.
 - Change the behavior of the "Add Another Option" button on the `PollsBuilderViewController` so that it can be pressed even if the keyboard is up, rather than needing to press to dismiss the keyboard and then again to press the button.
 - Fix the timer issues for good. 
 - Fix options for multiple choice in `PollsBuilderViewController` so that the trash icon displays correctly on all the options depending on the number of options there.

### DIFF
--- a/Clicker/ViewControllers/PollsDateViewController.swift
+++ b/Clicker/ViewControllers/PollsDateViewController.swift
@@ -10,8 +10,8 @@ import IGListKit
 import Presentr
 import UIKit
 
-protocol Reloadable {
-    func shouldReloadData()
+protocol PollsDateViewControllerDelegate {
+    func pollsDateViewControllerDidDeleteSession()
 }
 
 class PollsDateViewController: UIViewController {
@@ -30,7 +30,7 @@ class PollsDateViewController: UIViewController {
     var session: Session!
     var pollsDateArray: [PollsDateModel]!
     var numberOfPeople: Int = 0
-    var delegate: Reloadable!
+    var delegate: PollsDateViewControllerDelegate!
     
     // MARK: - Constants
     let countLabelWidth: CGFloat = 42.0
@@ -53,7 +53,7 @@ class PollsDateViewController: UIViewController {
         self.socket = Socket(id: "\(session.id)", userRole: userRole, delegate: self)
     }
     
-    func configure(with delegate: Reloadable) {
+    func configure(with delegate: PollsDateViewControllerDelegate) {
         self.delegate = delegate
     }
     
@@ -142,7 +142,7 @@ class PollsDateViewController: UIViewController {
         if pollsDateArray.isEmpty && session.name == session.code {
             DeleteSession(id: session.id).make()
                 .done {
-                    self.delegate.shouldReloadData()
+                    self.delegate.pollsDateViewControllerDidDeleteSession()
                 }
                 .catch { error in
                     print(error)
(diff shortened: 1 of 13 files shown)


Clicker/ViewControllers/PollsDateViewController.swift
```diff
@@ -128,20 +128,13 @@ class PollsDateViewController: UIViewController {
     
     @objc func goBack() {
         socket.socket.disconnect()
+        self.navigationController?.setNavigationBarHidden(true, animated: false)
+        self.navigationController?.popViewController(animated: true)
```
kevinchan159: do we want this to always happen? I think that if you pop the view controller before deleting, the home screen will still show the deleted poll.

Clicker/Views/MultipleChoice/AddMoreOptionCell.swift
```diff
@@ -27,8 +33,17 @@ class AddMoreOptionCell: UICollectionViewCell {
         setupViews()
     }
     
+    func configure(with: PollBuilderMCOptionSectionControllerDelegate) {
```
kevinchan159: rename this to `func configure(with delegate: PollBuilderMCOptionSectionControllerDelegate)`

Clicker/Views/PollBuilder/MCPollBuilderView.swift
```diff
@@ -123,9 +124,16 @@ class MCPollBuilderView: UIView, UITextFieldDelegate {
...
         }
         mcOptionModels.append(PollBuilderMCOptionModel(type: .addOption))
+        updateTotalOptions()
         adapter.performUpdates(animated: false, completion: nil)
     }
     
+    func updateTotalOptions() {
+        for opt in mcOptionModels {
+            opt.totalOptions = mcOptionModels.count - 1
+        }
```
kevinchan159: just a style choice but I like using `forEach` here

Clicker/Views/PollBuilder/MCPollBuilderView.swift
```diff
@@ -168,18 +176,19 @@ extension MCPollBuilderView: ListAdapterDataSource {
...
 extension MCPollBuilderView: PollBuilderMCOptionSectionControllerDelegate {
     
-    func pollBuilderSectionControllerShouldAddOption(sectionController: PollBuilderMCOptionSectionController) {
+    func pollBuilderSectionControllerShouldAddOption() {
         if mcOptionModels.count >= 27 { return }
         let newMCOptionModel = PollBuilderMCOptionModel(type: .newOption(option: "", index: mcOptionModels.count - 1))
         mcOptionModels.insert(newMCOptionModel, at: mcOptionModels.count - 1)
-        adapter.performUpdates(animated: false, completion: nil)
+        updateTotalOptions()
+        adapter.reloadData(completion: nil)
```
kevinchan159: why are we using `adapter.relaodData` instead of `adapter.performUpdates`

Clicker/ViewControllers/PollsDateViewController.swift
jschluger: I did this because if you need to delete the poll then there is a noticeable wait for the user before they see the app go back. Although, you are right that this way the home screen will show the deleted poll.

Clicker/ViewControllers/PollsDateViewController.swift
```diff
@@ -26,6 +30,7 @@ class PollsDateViewController: UIViewController {
     var session: Session!
     var pollsDateArray: [PollsDateModel]!
     var numberOfPeople: Int = 0
+    var delegate: Reloadable!
```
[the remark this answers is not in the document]
jschluger: Sure.

PR #197, opened 2018-10-12
Add latest activity labels for groups and other bug fixes

matthewcoufal:
simulator screen shot - iphone 8 plus - 2018-10-12 at 16 49 28


Clicker/Views/Cards/MCResultCell.swift
```diff
@@ -62,6 +62,8 @@ class MCResultCell: UICollectionViewCell {
     }
     
     override func updateConstraints() {
+        print("update constraints")
```
kevinchan159: can you remove this print

PR #199, opened 2018-10-12
Jack/debugging

jschluger:
Fixing more bugs from the hit list.


Clicker/ViewControllers/PollBuilderViewController+Extension.swift
```diff
@@ -10,6 +10,12 @@ import Foundation
 
 extension PollBuilderViewController: PollBuilderViewDelegate {
     
+    func ignoreNextKeyboardHiding() {
+        if isKeyboardShown {
+            shouldIgnoreNextKeyboardHiding = true
+        }
```
kevinchan159: can shorten this to `shouldIgnoreNextKeyboardHiding = isKeyboardShown`

Clicker/ViewControllers/PollsViewController.swift
```diff
@@ -67,10 +70,14 @@ class PollsViewController: UIViewController {
         
         setupViews()
         setupConstraints()
+        //setupGradient()
```
kevinchan159: can you remove this if we're not using it

Clicker/Views/PollBuilder/MCPollBuilderView.swift
```diff
@@ -181,7 +185,16 @@ extension MCPollBuilderView: PollBuilderMCOptionSectionControllerDelegate {
         let newMCOptionModel = PollBuilderMCOptionModel(type: .newOption(option: "", index: mcOptionModels.count - 1))
         mcOptionModels.insert(newMCOptionModel, at: mcOptionModels.count - 1)
         updateTotalOptions()
-        adapter.reloadData(completion: nil)
+        pollBuilderDelegate.ignoreNextKeyboardHiding()
+        adapter.reloadData { _ in
+            let index = IndexPath(item: 0, section: self.mcOptionModels.count - 2)
```
kevinchan159: what does `self.mcOptionModels.count - 2` refer to?

jschluger: This is because the last cell (at `mcOptionModels.count - 1`) is the add options button, so we want to select the one before that, the last cell for typing an option.

I'll leave a comment in the code to clarify.

PR #214, opened 2018-10-23
More bug fixes

matthewcoufal:
• When there are no answers on a free response question, it should say “No responses yet” in the answers area
• Semantics of latest activity: if a group/poll is modified at 11:59 PM, then at 12:01 AM it should say "1 day ago" instead of "< 1 hr ago"
• Move left of question label and miscellaneous icon to the right (refer to jason’s post in pod-pollo)
• Move total number of votes to the left (refer to jason’s post in pod-pollo)
• Fix count indicator label so that no spaces between “/” and less padding
• Margin for polls should be 18px

UPDATE: 
Waiting to hear back from Yuna as to whether or not the task below is a backend issue. It doesn't seem to be a problem on the frontend.
• Editing a draft does not save the edits when you tap “Save as Draft”

simulator screen shot - iphone 8 plus - 2018-10-22 at 21 34 46


Clicker/Extensions/Utils.swift
```diff
@@ -103,6 +103,9 @@ func calculatePollOptionsCellHeight(for pollOptionsModel: PollOptionsModel) -> C
         optionHeight = LayoutConstants.mcOptionCellHeight
     case .frOption(let frOptionModels):
         optionModels = frOptionModels
+        if optionModels.count == 0 {
```
AAAstorga:
```suggestion
        if optionModels.isEmpty {
```

Clicker/Quarks/SessionQuark.swift
```diff
@@ -147,21 +147,23 @@ struct GetPollSessions: ClickerQuark {
...
-        } else if today.days(from: latestActivityDate) < 7 {
-            let suffix: String = today.days(from: latestActivityDate) == 1 ? "day" : "days"
-            latestActivity += "\(today.days(from: latestActivityDate)) \(suffix) ago"
         } else {
-            let formatter: DateFormatter = DateFormatter()
-            formatter.dateStyle = .medium
-            formatter.timeStyle = .none
-            latestActivity += String(formatter.string(from: latestActivityDate).split(separator: ",")[0])
+            if today.days(from: latestActivityDate) < 7 {
+                let suffix: String = today.days(from: latestActivityDate) == 1 ? "day" : "days"
```
AAAstorga:
```suggestion
                let suffix = today.days(from: latestActivityDate) == 1 ? "day" : "days"
```

Clicker/ViewControllers/CardController.swift
```diff
@@ -123,13 +123,13 @@ class CardController: UIViewController {
         countLabelBackgroundView.snp.makeConstraints { make in
             make.top.equalTo(view.safeAreaLayoutGuide.snp.top).offset(countLabelBackgroundViewTopPadding)
             make.centerX.equalToSuperview()
-            make.width.equalTo(countLabelWidth + countLabelHorizontalPadding * 2)
+            make.width.equalTo(countLabelWidth)
             make.height.equalTo(countLabelHeight)
         }
         
         countLabel.snp.makeConstraints { make in
             make.center.equalTo(countLabelBackgroundView.snp.center)
```
AAAstorga:
```suggestion
            make.center.equalTo(countLabelBackgroundView)
```
Snapkit is cool bc it will infer the `.snp.cneter` on `countLabelBackgroundView`

Clicker/Views/Cards/NoResponsesCell.swift
```diff
@@ -0,0 +1,53 @@
...
+        noResponsesLabel = UILabel()
+        noResponsesLabel.text = noResponsesText
+        noResponsesLabel.textColor = .clickerGrey2
+        noResponsesLabel.font = ._12MediumFont
+        noResponsesLabel.textAlignment = .center
+        contentView.addSubview(noResponsesLabel)
+        
+    }
+    
+    override func updateConstraints() {
```
AAAstorga: is there a reason you override this?

According to [Apple docs](https://developer.apple.com/documentation/uikit/uiview/1622512-updateconstraints) "You should only override this method when changing constraints in place is too slow, or when a view is producing a number of redundant changes.". This method will be called a bunch of times, so you are adding a ton of constraints

Clicker/Views/Cards/NoResponsesCell.swift
```diff
@@ -0,0 +1,53 @@
...
+        
+        noResponsesLabel.snp.makeConstraints { make in
+            make.width.equalToSuperview()
```
AAAstorga: `make.width.center.equalToSuperview()`

Clicker/Views/Cards/PollOptionsCell.swift
```diff
@@ -122,20 +122,29 @@ extension PollOptionsCell: ListAdapterDataSource {
...
+            return models
         case .mcChoice(let mcChoiceModels):
+            models.append(topSpaceModel)
             models.append(contentsOf: mcChoiceModels)
-            break
+            models.append(bottomSpaceModel)
+            return models
         case .frOption(let frOptionModels):
-            models.append(contentsOf: frOptionModels)
+            if frOptionModels.count > 0 {
```
AAAstorga:
```suggestion
            if !frOptionModels.isEmpty {
```

Clicker/Views/Cards/PollOptionsCell.swift
```diff
@@ -146,7 +155,15 @@ extension PollOptionsCell: ListAdapterDataSource {
...
         } else {
-            return SpaceSectionController()
+            guard let pollOptionsModel = pollOptionsModel else { return ListSectionController() }
+            switch pollOptionsModel.type {
+            case .mcResult(_):
+                return SpaceSectionController(noResponses: false)
+            case .mcChoice(_):
+                return SpaceSectionController(noResponses: false)
+            case .frOption(let models):
+                return SpaceSectionController(noResponses: models.count == 0)
```
AAAstorga:
```suggestion
                return SpaceSectionController(noResponses: models.count.isEmpty)
```

Clicker/Views/Cards/NoResponsesCell.swift
matthewcoufal: That makes perfect sense. Was wondering about this actually. Thanks!

Clicker/SectionControllers/SpaceSectionController.swift
```diff
@@ -21,6 +27,12 @@ class SpaceSectionController: ListSectionController {
     }
     
     override func cellForItem(at index: Int) -> UICollectionViewCell {
+        if noResponses {
+            let cell = collectionContext?.dequeueReusableCell(of: NoResponsesCell.self, for: self, at: index) as! NoResponsesCell
+            cell.setNeedsUpdateConstraints()
+            cell.configure(with: spaceModel.space)
```
kevinchan159: You should call `configure` before calling `setNeedsUpdateConstraints`. The reason we put constraints in `updateConstraints` is if we need to pass in some data to the cell that will be used in making our constraints.

Clicker/Views/PollBuilder/MCPollBuilderView.swift
```diff
@@ -195,6 +195,7 @@ extension MCPollBuilderView: PollBuilderMCOptionSectionControllerDelegate {
     }
     
     func pollBuilderSectionControllerDidUpdateOption(option: String, index: Int) {
+        print("did update option: \(option)")
```
kevinchan159: Remove this print statement

PR #217, opened 2018-10-26
More layout bug fixes

matthewcoufal:
Ignore the expo settings and packager-info json files. Accidentally created those when trying to run a different project in this directory - I just deleted them.

- “Join” button text should be 50% opacity when button is disabled
- When displaying latest activity time stamp, currently we calculate # days ago by hours, but a better way would be to compare the calendar date of the last activity to todays date. 
- Overflow card is a different length for multi-choice VS free response, should be same length


Clicker/Quarks/DraftQuark.swift
```diff
@@ -106,7 +106,6 @@ struct UpdateDraft: ClickerQuark {
     }
     var parameters: Parameters {
         return [
-            "id": id,
```
kevinchan159: why are we removing this?

Clicker/SectionControllers/CardSectionControllers/FRInputSectionController 2.swift
```diff
@@ -0,0 +1,54 @@
+//
```
kevinchan159: is this a duplicate file?

Clicker/SectionControllers/CardSectionControllers/HamburgerCardSectionController 2.swift
```diff
@@ -0,0 +1,34 @@
+//
```
kevinchan159: duplicate file?

Clicker/SectionControllers/PollOptionsCellSectionControllers/FROptionSectionController.swift
```diff
@@ -36,8 +36,7 @@ class FROptionSectionController: ListSectionController {
         guard let containerSize = collectionContext?.containerSize else {
             return .zero
         }
-        let cellHeight = calculateFROptionCellHeight(for: frOptionModel)
-        return CGSize(width: containerSize.width, height: cellHeight)
+        return CGSize(width: containerSize.width, height: LayoutConstants.frOptionCellHeight)
```
kevinchan159: Is `frOptionCellHeight` going to be the same as the cell heights for the multiple choice cells?

Clicker/Quarks/DraftQuark.swift
matthewcoufal: Yuna told me that we don't need to specify the id in that parameters variable, that the request will still work without it.

PR #236, opened 2019-02-13
Testflight

AAAstorga:
I added a new way to manage secrets. I will send the env-vars.sh. You can delete Keys.plist because it is not needed anymore. 

We will now use Sourcery(https://github.com/krzysztofzablocki/Sourcery). It will generate a `keys.generated.swift` file on build time, thus our secrets are safe.

### DIFF
--- a/Sourcery/keys.stencil
+++ b/Sourcery/keys.stencil
@@ -1,0 +1,18 @@
+enum Keys: String {
+    case apiURL = "{{ argument.apiURL }}"
+    case apiDevURL = "{{ argument.apiDevURL }}"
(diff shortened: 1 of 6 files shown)


Sourcery/keys.stencil
```diff
@@ -0,0 +1,18 @@
...
+    }    
+    
+    static var hostURL: Keys {
+        #if DEV_SERVER
+        return Keys.apiDevURL
+        #else
+        return Keys.apiURL
+        #endif
+    }
+
```
kevinchan159: nit: delete line

PR #237, opened 2019-02-13
Disable button

mathewscullin:
- Set isEnabled boolean to false for peopleButton in the PollStateViewController


Clicker/ViewControllers/PollsDateViewController.swift
```diff
@@ -104,6 +104,7 @@ class PollsDateViewController: UIViewController {
         self.navigationItem.leftBarButtonItem = UIBarButtonItem(image: backImage, style: .done, target: self, action: #selector(goBack))
         
         peopleButton = UIButton()
+        peopleButton.isEnabled = false
```
mindylou: nice job!

PR #241, opened 2019-02-14
Austin/swiftlint

AAAstorga:
trying to see if the CI does swiftlint and reports issues back to PR


Clicker/Views/Cards/CardCell.swift
```diff
@@ -31,7 +31,7 @@ class CardCell: UICollectionViewCell {
     var timerLabel: UILabel!
     
     // MARK: - Data vars
-    var delegate: CardCellDelegate!
+    weak var delegate: CardCellDelegate!
```
mindylou: if these are `weak` should they be optional? when i refactored the delegates there was a lot of problems with making these optional tho bc of how often they're used (+ assumed to be there)

Clicker/Views/OptionsView.swift
```diff
@@ -19,9 +19,9 @@ class OptionsView: UIView, UICollectionViewDataSource, UICollectionViewDelegate,
     var options: [String]!
     var sliderBar: UIView!
     var sliderBarLeftConstraint: NSLayoutConstraint!
-    var sliderBarDelegate: SliderBarDelegate!
+    weak var sliderBarDelegate: SliderBarDelegate!
```
mindylou: make this optional

Clicker/Views/Cards/PollOptionsCell.swift
```diff
@@ -25,7 +25,7 @@ class PollOptionsCell: UICollectionViewCell, UIScrollViewDelegate {
     var arrowView: ArrowView!
     
     // MARK: - Data vars
-    var delegate: PollOptionsCellDelegate!
+    weak var delegate: PollOptionsCellDelegate!
```
mindylou: this should be optional too

Gemfile
```diff
@@ -1,3 +1,7 @@
-source "https://rubygems.org"
+source 'https://rubygems.org'
 
-gem "fastlane"
```
mindylou: wait if we don't have fastlane can we still use bitrise?

PR #244, opened 2019-02-16
Make arrow image user role sensitive

mathewscullin:
Arrow Image within NavigationTitleView now only display when the userRole is of type .admin.


Clicker/Views/Cards/NavigationTitleView.swift
```diff
@@ -61,11 +61,11 @@ class NavigationTitleView: UIView {
         addSubview(navigationButton)
     }
     
-    func configure(primaryText: String, secondaryText: String, delegate: NavigationTitleViewDelegate? = nil) {
+    func configure(primaryText: String, secondaryText: String, userRole: UserRole? = nil, delegate: NavigationTitleViewDelegate? = nil) {
```
kevinchan159: will userRole ever be `nil`? if not, then we don't need to make it an optional param

mathewscullin: The reason for it being optional is that configure for NavigationTitleView was being called in other parts of the CodeBase where userRole wasn't in that view controller.

kevinchan159: ah ok gotcha

lucyxubroad:
Your changes look good to me! Although I'm not really sure what these files are? 

`Podfile.lock`
`Pods/Manifest.lock`
`Pods/Pods.xcodeproj/project.pbxproj`

Are the changes intended?

Podfile.lock
```diff
@@ -120,7 +120,7 @@ SPEC CHECKSUMS:
   Google: 98da4b1a60ed0beb80a290387a28333406a1eb90
   GoogleSignIn: d9ef55b10f0aa401a5de2747f59b725e4b9732ac
   GoogleToolboxForMac: 2501e2ad72a52eb3dfe7bd9aee7dad11b858bd20
-  GTMOAuth2: e8b6512c896235149df975c41d9a36c868ab7fba
+  GTMOAuth2: c77fe325e4acd453837e72d91e3b5f13116857b2
```
mindylou: these are all pod conflicts, so when you merge with master you might have to delete `Podfile.lock` and the entire `Pods/` file... @AAAstorga the downsides of committing pods 😢

PR #255, opened 2019-03-06
Date fix

mindylou:
- Adjusts date of sessions to be with respect to user phone's time zone (not UTC) and by current day
- Redoes logic for displaying sorted polls to act like a stack (backend returns to them us newest first)
- Update date vars from backend to be parsed from backend as Unix timestamps (and then as a computed var `dateValue`) rather than formatted date string
- Looking for feedback regarding `dateValue` and storing `date` as a String timestamp ⏲

### DIFF
--- a/Clicker/Models/ListDiffableModels/PollsDateModel.swift
+++ b/Clicker/Models/ListDiffableModels/PollsDateModel.swift
@@ -21,20 +21,33 @@ struct PollsResponse: Codable {
 }
 
 struct GetSortedPollsResponse: Codable {
-    
+
+    // Seconds since 1970
     var date: String
     var polls: [PollsResponse]
-    
+
+    /// Swift Date representation of `date` sent from backend
+    lazy var dateValue: Date = {
+        return convertUnixStringToDate(date)
+    }()
+
 }
 
 class PollsDateModel: Codable {
-    
+
+    // Seconds since 1970
     var date: String
     var polls: [Poll]
     let identifier = UUID().uuidString
+
+    /// Swift Date representation of `date` sent from backend
+    lazy var dateValue: Date = { [weak self] in
+        guard let `self` = self else { return Date() }
+        return convertUnixStringToDate(self.date)
+    }()
     
     init(date: String, polls: [Poll]) {
-        self.date = date.trim()
+        self.date = date
         self.polls = polls
     }
     
--- a/Clicker/ViewControllers/PollsDateViewController+Extension.swift
+++ b/Clicker/ViewControllers/PollsDateViewController+Extension.swift
@@ -17,8 +17,7 @@ extension PollsDateViewController: ListAdapterDataSource {
             let type: EmptyStateType = .cardController(userRole: userRole)
             return [EmptyStateModel(type: type)]
         }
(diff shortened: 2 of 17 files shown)


Clicker/ViewControllers/PollsDateViewController+Extension.swift
```diff
@@ -97,7 +96,7 @@ extension PollsDateViewController: PollBuilderViewControllerDelegate {
         let newPoll = Poll(text: text, questionType: type, options: options, results: pollResults, state: state, correctAnswer: correctAnswer)
         appendPoll(poll: newPoll)
         adapter.performUpdates(animated: false, completion: nil)
-        if let lastPollsDateModel = pollsDateArray.last {
+        if let lastPollsDateModel = pollsDateArray.first {
```
kevinchan159: rename to `firstPollsDateModel`

Clicker/ViewControllers/PollsDateViewController+Extension.swift
```diff
@@ -167,10 +166,9 @@ extension PollsDateViewController: SocketDelegate {
     }
     
     func pollStarted(_ poll: Poll, userRole: UserRole) {
-        if let lastPollsDateModel = pollsDateArray.last {
-            if lastPollsDateModel.polls.contains(where: { otherPoll -> Bool in
-                return otherPoll.id == poll.id
-            }) || lastPollsDateModel.polls.last?.state == .live { return }
+        if let lastPollsDateModel = pollsDateArray.first {
+            if lastPollsDateModel.polls.contains(where: { $0.id == poll.id })
+                || lastPollsDateModel.polls.first?.state == .live { return }
```
kevinchan159: same comment as above

Clicker/Models/ListDiffableModels/PollsDateModel.swift
```diff
@@ -21,20 +21,36 @@ struct PollsResponse: Codable {
...
     var date: String
     var polls: [Poll]
     let identifier = UUID().uuidString
+
+    /// Swift Date representation of `date` sent from backend
+    lazy var dateValue: Date = { [weak self] in
+        guard let `self` = self, let dateAsDouble = Double(self.date) else { return Date() }
+        let unixDate = Date(timeIntervalSince1970: dateAsDouble)
+        return unixDate
+    }()
```
kevinchan159: since we do this twice, can we just make it a public helper function instead

Clicker/ViewControllers/PollsDateViewController+Extension.swift
```diff
@@ -263,18 +261,18 @@ extension PollsDateViewController: SocketDelegate {
     }
     
     func appendPoll(poll: Poll) {
-        let todaysDate = getTodaysDate()
-        if let lastPollDateModel = pollsDateArray.last {
-            if lastPollDateModel.date == todaysDate {
+        let todaysDate = Date()
+        if let lastPollDateModel = pollsDateArray.first {
+            if lastPollDateModel.dateValue.isSameDay(as: todaysDate) {
```
kevinchan159: same comment

PR #260, opened 2019-03-10
Merge latest changes into Testflight

mindylou:
- Because of backend changes with date display, the parsing for polls grouped by date is off/looks bad
- Ability to un-upvote


AAAstorga:
Make sure to change testflight notes!

PR #266, opened 2019-03-24
Prompt user ratings every 10 launches

lucyxubroad:
Prompt users for ratings every ten launches of app. 

Files added: 
- Ratings.swift: Compartmentalize code for prompting user ratings. Every time a user launches Pollo, increment "NUMBER_OF_LAUNCHES" by 1 and prompt for rating after 10 launches. Reset "NUMBER_OF_LAUNCHES" to 0 after prompt. 

Question: Where should "Ratings.swift" go?

Screenshot:
Screen Shot 2019-03-24 at 3 32 36 PM

### DIFF
--- a/Clicker/Models/User.swift
+++ b/Clicker/Models/User.swift
@@ -57,7 +57,6 @@ struct UserSession: Codable {
     let refreshToken: String
     let sessionExpiration: String
 
-    
     init(accessToken: String, refreshToken: String, sessionExpiration: String, isActive: Bool) {
         self.accessToken = accessToken
         self.isActive = isActive
--- a/Clicker/Models/User.swift
+++ b/Clicker/Models/User.swift
@@ -56,6 +56,7 @@ struct UserSession: Codable {
     let isActive: Bool
     let refreshToken: String
     let sessionExpiration: String
+
     
     init(accessToken: String, refreshToken: String, sessionExpiration: String, isActive: Bool) {
         self.accessToken = accessToken
--- a/Clicker/Supporting/Ratings.swift
+++ b/Clicker/Supporting/Ratings.swift
@@ -11,7 +11,7 @@ import StoreKit
 
 class Ratings {
     
-    static let launchThreshold = 10
+    static let launchThreshold = 1
     
     class func updateNumAppLaunches() {
         let numLaunches = getNumAppLaunches() + 1
--- a/Clicker/Supporting/Ratings.swift
+++ b/Clicker/Supporting/Ratings.swift
@@ -11,26 +11,28 @@ import StoreKit
 
 class Ratings {
     
-    static let launchThreshold = 20
-    static let NUMBER_OF_LAUNCHES = "NUMBER_OF_LAUNCHES"
+    static let shared = Ratings()
+    let launchThreshold = 20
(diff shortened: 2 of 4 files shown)


Clicker/Supporting/Ratings.swift
```diff
@@ -0,0 +1,43 @@
...
+//  Created by Lucy Xu on 3/24/19.
+//  Copyright © 2019 CornellAppDev. All rights reserved.
+//
+
+import Foundation
+import StoreKit
+
+class Ratings {
+    
+    static let launchThreshold = 10
```
kevinchan159: i think it would be better to make this a singleton so that we don't actually instantiate an instance of `Ratings`. so do something like
```
static let shared = Ratings()

private init() {}
```
You would also remove all the `class` declarations before every function and then anywhere you want to call Ratings you can do `Ratings.shared.updateNumAppLaunches()`

Clicker/Supporting/Ratings.swift
```diff
@@ -0,0 +1,43 @@
...
+    }
+    
+    class func getNumAppLaunches() -> Int {
+        if let numLaunches = UserDefaults.standard.value(forKey: "NUMBER_OF_LAUNCHES") as? Int {
+            return numLaunches
+        }
+        return 0
+    }
+    
+    class func shouldPromptReview() -> Bool{
```
kevinchan159: space after `Bool` and `{`

Clicker/Models/User.swift
```diff
@@ -54,10 +54,10 @@ struct UserSession: Codable {
     
     let accessToken: String
     let refreshToken: String
-    let sessionExpiration: String
+    let sessionExpiration: Int
```
kevinchan159: can we alphabetize these

Clicker/ViewControllers/PollsViewController.swift
```diff
@@ -72,13 +73,16 @@ class PollsViewController: UIViewController {
...
+        
         view.backgroundColor = .clickerGrey8
 
         NotificationCenter.default.addObserver(self, selector: #selector(keyboardWillShow), name: NSNotification.Name.UIKeyboardWillShow, object: nil)
         NotificationCenter.default.addObserver(self, selector: #selector(keyboardWillHide), name: NSNotification.Name.UIKeyboardWillHide, object: nil)
         
+        promptUserReview()
         setupViews()
         setupConstraints()
+
```
kevinchan159: delete line

kevinchan159:
Can you also make sure you have the latest master and then run `git merge master` on your branch. I think my alphabetization stuff might not be here

Clicker/Supporting/Ratings.swift
```diff
@@ -0,0 +1,43 @@
...
+        let numLaunches = getNumAppLaunches()
+        return numLaunches >= launchThreshold
+    }
+    
+    class func promptReview(){
+        if #available(iOS 10.3, *)  {
+            if shouldPromptReview() {
```
mindylou: chain these together so `#available... && shouldPromptReview()`

mindylou:
I think `Ratings.swift` should probably go in "Supporting" so you put it in the right place!

Clicker/Supporting/Ratings.swift
mindylou: ^^

PR #277, opened 2019-04-25
Implement press enter to join poll and incorrect code handling

mindylou:
- Pressing "enter" on the keyboard allows you to join a poll (if 6 characters are entered)
- Fixes error with decoding so that error case w/incorrect poll code is now handled

### DIFF
--- a/Clicker/ViewControllers/PollsViewController+Extension.swift
+++ b/Clicker/ViewControllers/PollsViewController+Extension.swift
@@ -57,7 +57,7 @@ extension PollsViewController: PollsCellDelegate {
         isOpeningGroup = true
         joinSessionWithIdAndCode(id: session.id, code: session.code).chained { sessionResponse -> Future<Response<[PollsDateModel]>> in
             let session = sessionResponse.data
-            return self.getSortedPolls(with: session.id)
+            return self.getSortedPolls(with: session.id).decode()
         }.observe { [weak self] result in
             guard let `self` = self else { return }
             DispatchQueue.main.async {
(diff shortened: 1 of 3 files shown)


Clicker/ViewControllers/PollsViewController+Extension.swift
```diff
@@ -162,7 +162,10 @@ extension PollsViewController: SliderBarDelegate {
 extension PollsViewController: UITextFieldDelegate {
 
     func textFieldShouldReturn(_ textField: UITextField) -> Bool {
-        self.view.endEditing(true)
+        if let text = textField.text, text.count == 6 { // Valid code length
```
matthewcoufal: Can we make `6` into a class variable at the top of `PollsViewController` like `let validCodeLength = 6`? Just in case we ever decide to change the valid code length.

Clicker/ViewControllers/PollsViewController.swift
```diff
@@ -332,29 +332,44 @@ class PollsViewController: UIViewController {
...
+            // Fail out of get sorted polls call
+            return self.getSortedPolls(with: -1)
         }.observe { [weak self] result in
-            guard let `self` = self, let session = self.session else { return }
-            DispatchQueue.main.async {
-                switch result {
-                case .value(let pollsResponse):
-                    
+            guard let `self` = self else { return }
+            let jsonDecoder = JSONDecoder()
```
lucyxubroad: Lol, I'm not entirely sure about this one but in Socket, I think jsonDecoder became a defined variable at the top - which I think we can also do here to remove duplicate code? I might be wrong though.

```
var jsonDecoder: JSONDecoder = {
        let decoder = JSONDecoder()
        decoder.dateDecodingStrategy = .secondsSince1970
        return decoder
    }()
```

mindylou: ooo great idea

Clicker/ViewControllers/PollsViewController+Extension.swift
mindylou: i'm gonna actually add it to `IntegerConstants` in case we use it anywhere else!

PR #279, opened 2019-04-28
Implement ability to activate/deactivate filter

mindylou:
- Part one of filtering profanity (smaller PR)
- Implement ability to activate/deactivate filtering in group controls
- Implement barebones socket delegate functions (to be implemented in another PR)

### DIFF
--- a/Clicker/Extensions/UITextField+Shared.swift
+++ b/Clicker/Extensions/UITextField+Shared.swift
@@ -14,8 +14,8 @@ extension UITextField {
         guard let text = self.text else { return }
         let nsString = text as NSString
         let attributedString = NSMutableAttributedString(string: text)
-        for word in words {
-            let range = nsString.range(of: word)
+        words.forEach {
+            let range = nsString.range(of: $0)
             attributedString.addAttribute(.foregroundColor, value: UIColor.red, range: range)
         }
         attributedText = attributedString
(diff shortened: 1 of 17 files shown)


Clicker/Extensions/UITextField+Shared.swift
```diff
@@ -0,0 +1,24 @@
...
+
+import Foundation
+
+extension UITextField {
+
+    func updateTextToDisplayProfanity(for words: [String]) {
+        guard let text = self.text else { return }
+        let nsString = text as NSString
+        let attributedString = NSMutableAttributedString(string: text)
+        for word in words {
```
kevinchan159: pref to use `forEach` loop

Clicker/ViewControllers/GroupControlsViewController.swift
```diff
@@ -238,3 +250,30 @@ extension GroupControlsViewController: AttendanceViewControllerDelegate {
     }
 
 }
+
+// MARK: - PollSettingsSectionControllerDelegate
+extension GroupControlsViewController: PollSettingsSectionControllerDelegate {
+
+    func pollSettingsSectionControllerDidUpdate(_ sectionController: PollSettingsSectionController, to newValue: Bool) {
+        guard let pollsSettingModel = adapter.object(for: sectionController) as? PollsSettingModel else { return }
+        if pollsSettingModel.isEqual(toDiffableObject: filterSetting) {
```
kevinchan159: might be cleaner if you move this to the `guard let`

mindylou: eventually there's going to be more model toggles so each one would be in an if statement i think

PR #280, opened 2019-04-29
Dynamic Plus Button and Group Controls Bug Fix

mathewscullin:
Plus button no longer shows up when on the "Joined" side on PollsViewController.
Group Controls arrow bug fix.


Clicker/Views/OptionsView.swift
```diff
@@ -91,9 +94,12 @@ class OptionsView: UIView, UICollectionViewDataSource, UICollectionViewDelegate,
         let unselectedIndexPath: IndexPath
         if indexPath.row == 0 {
             unselectedIndexPath = IndexPath(item: 1, section: 0)
+            isJoined = true
```
mindylou: instead of this if else statement block
```
unselectedIndexPath = indexPath.row == 0 ? IndexPath(item: 1, section: 0) : IndexPath(item: 0, section: 0)
isJoined = indexPath.row == 0 
```

PR #300, opened 2019-09-23
Remove FR UI from PollBuilderViewController

mathewscullin:
Admins can no longer start a free-response Poll.
Eradicated ```frPollBuilder```, ```QuestionTypeDropDownView``` and its respective functions from PollBuilderViewController.

image


Pollo/ViewControllers/PollBuilderViewController.swift
```diff
@@ -367,36 +295,7 @@ class PollBuilderViewController: UIViewController {
...
-                        case .error(let error):
-                            print("error: ", error)
-                        }
-                    }
-                }
-            }
-            loadedFRDraft = nil
-            self.frPollBuilder.reset()
-        }
+        case .freeResponse: break }
```
kevinchan159: oh is this what you were asking about? the `}` should be on a newline

Pollo/ViewControllers/PollBuilderViewController.swift
```diff
@@ -417,34 +316,19 @@ class PollBuilderViewController: UIViewController {
             answerChoices = mcPollBuilder.getChoices()
             let question = loadedMCDraft?.text ?? ""
             delegate?.startPoll(text: question, type: .multipleChoice, options: mcPollBuilder.getOptions(), state: .live, answerChoices: answerChoices, correctAnswer: correctAnswer, shouldPopViewController: true)
-        case .freeResponse:
-            let question = loadedFRDraft?.text ?? ""
-            delegate?.startPoll(text: question, type: .freeResponse, options: [], state: .live, answerChoices: [], correctAnswer: nil, shouldPopViewController: true)
-        }
+        case .freeResponse: break }
```
kevinchan159: same

Pods/Pods.xcodeproj/project.pbxproj
```diff
@@ -6,3183 +6,3875 @@
 	objectVersion = 48;
 	objects = {
 
+/* Begin PBXAggregateTarget section */
```
jackthmp: @kevinchan159 are all these changes okay? I'm assuming they're just related to updating CocoaPods but I'm curious what will happen when this is merged.

kevinchan159: yeah I think its just from the cocoapod update so should be fine. i would double check that it runs on your computer before merging though @jackthmp

PR #302, opened 2019-09-29
Fix UITextField bug in CreateMCOptionCell

matthewcoufal:
This PR fixes a bug in `CreateMCOptionCell` where text field events would not fire and the trash icon would not be visible.

Here is what `PollBuilderViewController` looks like with these changes in place:

Simulator Screen Shot - iPhone 8 Plus - 2019-09-29 at 16 42 21


Pollo.xcodeproj/project.pbxproj
```diff
@@ -27,16 +28,12 @@
 		0CDAF0D321727D4B004DB6EA /* SpaceSectionController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0CDAF0D221727D4B004DB6EA /* SpaceSectionController.swift */; };
 		0CDAF0D521727DA3004DB6EA /* SpaceCell.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0CDAF0D421727DA3004DB6EA /* SpaceCell.swift */; };
 		0CF647832176970900535573 /* ArrowView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0CF647822176970900535573 /* ArrowView.swift */; };
-		20FDA08B232BF481006082B8 /* GoogleService-Info.plist in Resources */ = {isa = PBXBuildFile; fileRef = 20FDA08A232BF481006082B8 /* GoogleService-Info.plist */; };
```
jackthmp: Looks like of the free response files were dereferenced in the project file. Has this stuff already been merged in?

matthewcoufal: @jackthmp Yes. I forgot to add the `Pollo.xcodeproj/project.pbxproj` file in my last PR, so I removed the references in this PR so that the app will build properly.

PR #304, opened 2019-10-03
Fix Create Poll Button

mathewscullin:
Both `createPollButton` and the group controls arrow show up when they are meant to. The test plan was by navigating `pollsViewController`, `pollsDateViewController`, and `cardController` with a poll live and a poll ended. This requires ensuring `session.isLive` is correctly set across multiple ViewControllers. Additionally, in `pollsBuilderViewController` the header label now says `Create Question` and properly loads the title to the Poll.

### DIFF
--- a/Pollo/ViewControllers/CardController.swift
+++ b/Pollo/ViewControllers/CardController.swift
@@ -170,6 +170,7 @@ class CardController: UIViewController {
         
         if userRole == .admin {
             createPollButton = UIButton()
+            createPollButton.isHidden = session.isLive ?? false
             createPollButton.setImage(#imageLiteral(resourceName: "whiteCreatePoll"), for: .normal)
             createPollButton.addTarget(self, action: #selector(createPollBtnPressed), for: .touchUpInside)
             let createPollBarButton = UIBarButtonItem(customView: createPollButton)
--- a/Pollo/ViewControllers/PollBuilderViewController.swift
+++ b/Pollo/ViewControllers/PollBuilderViewController.swift
@@ -31,7 +31,7 @@ class PollBuilderViewController: UIViewController {
     var dimmingView: UIView!
     var exitButton: UIButton!
     var mcPollBuilder: MCPollBuilderView!
-    var multipleChoiceLabel: UILabel!
+    var createQuestionLabel: UILabel!
     var quizModeOverlayView: QuizModeOverlayView!
     var saveDraftButton: UIButton!
     var startPollButton: UIButton!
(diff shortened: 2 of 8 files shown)


Pollo/ViewControllers/PollsViewController+Extension.swift
```diff
@@ -199,7 +199,7 @@ extension PollsViewController: UIGestureRecognizerDelegate {
     
 }
 
-extension PollsViewController: PollsDateViewControllerDelegate {
+extension PollsViewController: PollsDateViewControllerDelegate {
```
kevinchan159: nit: delete whitespace at end of line

Pollo/ViewControllers/PollBuilderViewController.swift
```diff
@@ -304,15 +305,15 @@ class PollBuilderViewController: UIViewController {
...
+
         delegate?.showNavigationBar()
         dismiss(animated: true, completion: nil)
         hideKeyboard()
         
         switch questionType { 
         case .multipleChoice:
             answerChoices = mcPollBuilder.getChoices()
-            let question = loadedMCDraft?.text ?? ""
+            let question = mcPollBuilder.questionText ?? ""
```
jackthmp: Can you explain this change?

Pollo/ViewControllers/CardController.swift
```diff
@@ -170,6 +170,7 @@ class CardController: UIViewController {
         
         if userRole == .admin {
             createPollButton = UIButton()
+            createPollButton.isHidden = pollsDateModel.polls.last?.state == .live
```
jackthmp: should this be `= session.isLive` ?

Pollo/ViewControllers/PollBuilderViewController.swift
mathewscullin: For some reason, we were loading the text field from `loadedMCDraft` for a poll which means if the admin did not make a question from a draft, the question field they typed wasn't loaded into the poll. Now if an admin makes a question directly `mcPollBuilder.questionText` loads what they typed in that field and if they use a draft, `mcPollBuilder.questionText` also gets set with what `loadedMCDraft.text`, so it covers both cases.

Pollo/ViewControllers/CardController.swift
mathewscullin: For consistency, we could use it but within `cardController`, `pollsDateModel` holds the same data and is always correct. The fact that we are altering `session.isLive` made me only want to use it when it was a necessity.

PR #305, opened 2019-10-03
Update home screen icons and alert messages

matthewcoufal:
This PR updates the image asset for the ellipsis icon in `PollPreviewCell` and the alert message for a failure to join a group to reflect the current designs.

Here is what the new ellipsis icon looks like:

Simulator Screen Shot - iPhone 8 Plus - 2019-10-02 at 23 30 18

Here is what the new alert message looks like:

Simulator Screen Shot - iPhone 8 Plus - 2019-10-02 at 23 34 29


Pollo/ViewControllers/PollsViewController.swift
```diff
@@ -362,15 +362,14 @@ class PollsViewController: UIViewController {
             case .value(let data):
                 guard let pollsResponse = try? self.jsonDecoder.decode(Response<[GetSortedPollsResponse]>.self, from: data), pollsResponse.success else {
                     DispatchQueue.main.async {
-                        let alertController = self.createAlert(title: self.errorText, message: "Failed to join session with code \(code). Try again!")
+                        let alertController = self.createAlert(title: self.errorText, message: "We couldn't find a group with that code. Please try again.", actionTitle: "Okay")
```
mathewscullin: Seems like in Pollo some alert controllers save their message as a constant above while others have them in the `message:` parameter. What should we adhere to maintain uniformity?

matthewcoufal: @mathewscullin  I agree - just made the message a constant in `PollsViewController`.

PR #316, opened 2019-10-20
Change Create Session Method

mathewscullin:
## Overview
Admins create sessions now using a text field at the bottom only when on the created side of ```PollsViewController```. This change is based on user research and testing from the previous semester where users clicked the ```NewGroupButton``` to join a group rather than typing below. This design change aims to eliminate this confusion by removing ```NewGroupButton``` entirely.
## Changes Made

I made an additional `createSessionContainerView` that has a text field and button that handle all the functionality of starting a new group. `createSessionContainerView` and `joinSessionContainerView` switch depending on if the collectionView is on `Joined` or `Created`. I consequently removed `NewGroupButton` and `NameView.swift.`

## Test Coverage
I made new groups with this method and ensured that they worked as expected.

## Next Steps

The core functionality is complete in this PR, however, I will followup with another one that will animate the transition between the two container views, for a more fluid transition.

## Screenshots
<details>

  <summary>New Text Field</summary>

  Simulator Screen Shot - iPhone X - 2019-10-20 at 11 41 31

</details>


Pollo/ViewControllers/PollsViewController.swift
```diff
@@ -219,6 +243,24 @@ class PollsViewController: UIViewController {
...
         }
+
+        createSessionContainerView.snp.makeConstraints { make in
+            make.bottom.equalTo(view.safeAreaLayoutGuide.snp.bottom)
+            make.width.centerX.equalToSuperview()
+            make.height.equalTo(joinSessionContainerViewHeight)
+        }
+
+        createSessionButton.snp.makeConstraints { make in
+            make.width.equalTo(83.5)
```
CameronHamidi: Should this 83.5 be a constant somewhere?

Pollo/ViewControllers/PollsViewController.swift
```diff
@@ -502,6 +546,14 @@ class PollsViewController: UIViewController {
                 make.height.equalTo(joinSessionContainerViewHeight)
             }
             joinSessionContainerView.superview?.layoutIfNeeded()
+
+            createSessionContainerView.snp.remakeConstraints { make in
+
```
CameronHamidi: Delete newline

Pollo/ViewControllers/PollsViewController+Extension.swift
```diff
@@ -163,10 +168,10 @@ extension PollsViewController: SliderBarDelegate {
     func scrollToIndex(index: Int) {
         let indexPath = IndexPath(item: 0, section: index)
         pollsCollectionView.scrollToItem(at: indexPath, at: [], animated: true)
-    }
 
-    func changeNewGroupButton(status: Bool) {
-        newGroupButton.isHidden = status
+        self.bottomPaddingView.backgroundColor = index == 1 ? .lightGrey : .darkestGrey
+        self.joinSessionContainerView.isHidden = index == 1
+        self.createSessionContainerView.isHidden = index == 0
```
kevinchan159: can we make `index == 0` and `index == 1` into variables which describe what they represent?

Pollo/ViewControllers/PollsViewController.swift
```diff
@@ -323,34 +381,9 @@ class PollsViewController: UIViewController {
...
-        return networking(Endpoint.joinSessionWithCode(with: code))
-    }
-    
-    func getSortedPolls(with id: Int) -> Future<Data> {
-        return networking(Endpoint.getSortedPolls(with: id))
-    }
-    
     @objc func joinSession() {
         guard let code = codeTextField.text, code != "" else { return }
+        codeTextField.text = ""
```
kevinchan159: are we sure that we want to clear the textfield right away? I feel like it might be odd if we clear the textfield and the network request takes a while to respond.
as a side note, is there some loading indicator displayed when they click join?

mathewscullin: I will change this so that it is cleared later in the function once the request is successful. As for your followup question, there is no loading indicator, the request is very quick almost instantaneous.

Pollo/ViewControllers/PollsViewController+Extension.swift
mathewscullin: Yup!

Pollo/ViewControllers/PollsViewController.swift
```diff
@@ -51,6 +53,9 @@ class PollsViewController: UIViewController {
     let codeTextFieldHeight: CGFloat = 40
     let codeTextFieldHorizontalPadding: CGFloat = 12
     let codeTextFieldPlaceHolder = "Enter a group code..."
+    let createSessionTextFieldPlaceHolder = "Enter a new group name..."
+    let createSessionButtonTitle = "Create"
+    let createSessionButtownWidth: CGFloat = 83.5
```
jackthmp: Typo

Pollo/ViewControllers/PollsViewController.swift
jackthmp: But if your network connection is slow, we would want to have some sort of user feedback so they know that their group is being created in case the network request doesn't return right away. I'd suggest adding a loading indicator, maybe over the "Create" button?

Pollo/ViewControllers/PollsViewController.swift
```diff
@@ -406,22 +439,36 @@ class PollsViewController: UIViewController {
...
     
-    @objc func didStartTyping(_ textField: UITextField) {
+    @objc func didStartTypingCode(_ textField: UITextField) {
         if let text = textField.text {
             textField.text = text.uppercased()
             updateJoinSessionButton(canJoin: text.count == IntegerConstants.validCodeLength)
         }
     }
+
+    @objc func didStartTypingName(_ textField: UITextField) {
```
jackthmp: nit: can you rename this to something more descriptive? Like `didStartTypingGroupName`?

Issue #322, opened 2019-11-03
Implement Onboarding Flow for Creating a Quiz


jackthmp:
Designs on Zeplin.

PR #332, opened 2019-11-13
Implement new onboarding flow

matthewcoufal:
## Overview

This PR refactors `QuizModeOverlayView` into `OnboardingView`, which is the implementation of the new onboarding flow. It also adds a new `UserDefaults` key that is used to decide whether or not to display `OnboardingView` in `PollBuilderViewController`.

## Changes Made

The new onboarding flow contains different stages, each of which describes a feature of Pollo. I created an enum `OnboardingStage` to represent each of these stages. In `transitionButtonTapped` inside `OnboardingView` I switched on the current stage and set up animated transitions between stages with the help of `UIView.animate`.

## Test Coverage

I used multiple device simulators to run through each stage of the onboarding flow and make sure it corresponds to the current designs.

## Related PRs or Issues

https://github.com/cuappdev/pollo-ios/issues/322

## Screenshots

<details>

  <summary>Welcome Stage</summary>

  
  
Simulator Screen Shot - iPhone 11 - 2019-11-13 at 00 23 00

</details>
<details>

  <summary>Create Question Stage</summary>

  
  
Simulator Screen Shot - iPhone 11 - 2019-11-13 at 00 23 02

</details>
<details>

  <summary>Autofill Stage</summary>

  
  
Simulator Screen Shot - iPhone 11 - 2019-11-13 at 00 23 04

</details>
<details>

  <summary>Quiz Mode Stage</summary>

  
  
Simulator Screen Shot - iPhone 11 - 2019-11-13 at 00 23 06

</details>
<details>

  <summary>Save Draft Stage</summary>

  
  
Simulator Screen Shot - iPhone 11 - 2019-11-13 at 00 23 07

</details>
<details>

  <summary>Start Question Stage</summary>

  
  
Simulator Screen Shot - iPhone 11 - 2019-11-13 at 00 23 09

</details>
<details>

  <summary>Finish Stage</summary>

  
  
Simulator Screen Shot - iPhone 11 - 2019-11-13 at 00 23 10

</details>


Pollo/Views/PollBuilder/OnboardingView.swift
```diff
@@ -0,0 +1,524 @@
...
+    private let secondOptionLabel = UILabel()
+    private let secondOptionView = UIView()
+    private let transitionButton = UIButton()
+    private let welcomeContainerView = UIView()
+    private let welcomeDescriptionLabel = UILabel()
+    private let welcomeTitleLabel = UILabel()
+    
+    // MARK: - Data vars
+    var containerFrame: CGRect!
+    var isTransitioning = false
```
mathewscullin: Is `isTransitiong` to prevent skipping a stage if a user taps during an animation? If so, it doesn't seem to be used anywhere.

Pollo/Views/PollBuilder/OnboardingView.swift
```diff
@@ -0,0 +1,524 @@
...
+        if stage == .createQuestion {
+            return "To customize this question, type here and add additional options"
+        } else if stage == .autofillChoices {
+            return "To autofill choices A, B, ... , tap here to start live polling"
+        } else if stage == .quizMode {
+            return "To use quiz mode, select a correct answer"
+        } else if stage == .saveDraft {
+            return "To come back to this question, tap here"
+        } else if stage == .startQuestion {
+            return "To start polling now, tap here"
```
mathewscullin: With so many if else's here, should we use a switch instead and for `case .welcome` have it be the empty string?

jackthmp: +1

Pollo/Views/PollBuilder/OnboardingView.swift
```diff
@@ -0,0 +1,524 @@
...
+    
+    @objc func transitionButtonTapped() {
+        if !shouldTransition { return }
+        shouldTransition = false
+        switch stage {
+        case .welcome:
+            stage = .createQuestion
+            animatedLabel.text = getAnimatedLabelText()
+            UIView.animate(withDuration: animationDuration, animations: {
+                self.welcomeContainerView.alpha = 0
```
jackthmp: I'm curious if it would be easier if you combined all of the views into superviews for each "slide" of the onboarding flow. This would avoid having to individually change the alpha values for each label/view within the slide.

Pollo/Views/PollBuilder/OnboardingView.swift
matthewcoufal: Had a separate variable `shouldTransition` used for this purpose, but I'll change that one to `isTransitioning` because that makes more sense. Thanks for catching that!

Pollo/Views/PollBuilder/OnboardingView.swift
matthewcoufal: Great idea! I'll try that right now.

matthewcoufal: @jackthmp This is done for `welcomeContainerView` for the first slide, but I'm not able to combine the rest of the views into superviews because they are reused throughout the remaining slides.

jackthmp: Ahh I see, dropping this then

Pollo/Views/PollBuilder/OnboardingView.swift
```diff
@@ -0,0 +1,528 @@
...
+                self.buttonLabel.alpha = 0
+                self.animatedLabel.snp.remakeConstraints { remake in
+                remake.leading.equalTo(self.firstCorrectView.snp.trailing).offset(self.getAnimatedLabelLeadingOffset())
+                    remake.top.equalTo(self.firstCorrectView).offset(-self.animatedLabelBottomSmallOffset)
+                    remake.trailing.equalToSuperview().inset(self.getAnimatedLabelTrailingInset())
+                }
+                self.layoutIfNeeded()
+            }) { completed in
+                if completed {
+                    self.animatedLabel.text = self.getAnimatedLabelText()
```
jackthmp: The label's transition (text change and left alignment) looks pretty jumpy, especially with slow animations turned on. Generally, let's change the text _before_ animating for consistency with the other transitions.
Also, some of the jumpiness is likely because the label's frame is changing size and cuts off the text for a brief second, especially when it switches from right to left alignment:
image

Something you could try is to hardcode the size of the `animatedLabel`'s frame so that it doesn't resize every time the text changes content or alignment. Alternatively you can play around with using a UIView.transition every time you change animatedLabel to make it look smoother?
Ex:
`UIView.transition(with: animatedLabel, duration: 0.25, options: .transitionCrossDissolve, animations: {
...
}`

matthewcoufal: Sounds good. I'll try that. Thanks!

matthewcoufal: @jackthmp I made it so that the text changes before the animation occurs. Does this still look jumpy to you?

PR #337, opened 2019-11-21
Fix Result Bars

mathewscullin:
## Overview

Update `containerView` and `highlightView` in` MCResultCell` to reflect new designs. Additionally this PR solves the bug where an admin does not see the correct answer they set in `Card Cell` and the bug where if a student leaves and returns their submission if correct or incorrect is no longer reflects that. Lastly, the end poll button color bug has been fixed.

## Changes Made

Changed varying attributes to the respective UIElements depending on whether `UserRole` is `member` or `admin`. Lastly `correctAnswer` was added to the response data for `getSortedPolls` to solve the bugs.

## Test Coverage

Compared changes to Zeplin files and tested to ensure the bugs no longer persisted.

## Screenshots

<details>

  <summary>Student Side</summary>

  
  Simulator Screen Shot - iPhone Xʀ - 2019-11-20 at 22 30 51

</details>

<details>

  <summary>Admin Side</summary>

  
Simulator Screen Shot - iPhone Xʀ - 2019-11-20 at 22 31 00

</details>

### DIFF
--- a/Pollo/Views/Cards/MCResultCell.swift
+++ b/Pollo/Views/Cards/MCResultCell.swift
@@ -31,13 +31,14 @@ class MCResultCell: UICollectionViewCell {
     
     // MARK: - Constants
     let checkImageName = "correctanswer"
-    let containerViewBorderWidth: CGFloat = 0.6
+    let containerViewBorderWidth: CGFloat = 0.3
     let containerViewCornerRadius: CGFloat = 8
     let containerViewHeight: CGFloat = 46
     let containerViewTopPadding: CGFloat = 8
     let correctImageName = "correct"
     let dotViewBorderWidth: CGFloat = 2
     let dotViewLength: CGFloat = 23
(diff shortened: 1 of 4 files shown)


Pollo/Views/Cards/MCResultCell.swift
```diff
@@ -148,8 +152,11 @@ class MCResultCell: UICollectionViewCell {
         }
 
         highlightView.snp.makeConstraints { make in
-            make.leading.top.bottom.trailing.equalToSuperview()
+            make.leading.top.bottom.equalToSuperview()
+            let highlightViewMaxWidth = Float(contentView.bounds.width - horizontalPadding * 2)
```
jackthmp: I don't think you need `highlightViewMaxWidth`, because the max with will be the width of `containerView` (superView), right? Does it work if you just do `make.width.equalToSuperview().multipliedBy(percentSelected)`?

PR #341, opened 2019-12-10
Fixed duplicate poll by replacing active poll with new instance

jackthmp:
## Overview

When the Socket disconnects and reconnects, it returns a new instance of the currently active Poll on both the admin and student side. Currently, we just append the new poll to the end of `PollsDateModel.polls`, which resulted in duplicates.

## Changes Made

Added a check to check if the Socket-returned Poll already exists, and if so just replace that Poll in the `polls` list with the new instance of the poll.

## Test Coverage

Manual testing, built on my phone and put app in background and brought it back and the active poll no longer duplicates.

#311


Pollo/ViewControllers/CardController+Extension.swift
```diff
@@ -260,7 +260,12 @@ extension CardController: SocketDelegate {
             delegate?.pollStarted(poll, userRole: userRole)
             return
         }
+
+        if pollsDateModel.polls.contains(where: { $0.isEqual(toDiffableObject: poll)}) {
+            pollsDateModel.polls.removeLast()
```
kevinchan159: why do we `removeLast()`, shouldn't we remove the index of the duplicate poll?

jackthmp: i'm pretty sure the duplicate/active poll is always going to be the last poll in the list, but i guess it could be safer to remove at the index of the duplicate?

jackthmp: fixed with a filter!

Issue #335, opened 2019-11-17
Update API Spec Changes


jackthmp:
API Poll
`type` - removed

Draft
`options: string[]` - will no longer be empty for FR

Poll (Types)
`PollType` - type removed
`PollResult` - `letter` is no longer optional
`PollChoice` - `letter` is no longer optional

Poll (Fields)
`upvotes` - removed
`answerChoices: PollResult[]` - see `PollResult`
`answers: { googleID : PollChoice[] }`  - see `PollChoice`
`correctAnswer` - no behavioral changes, just won’t be empty string if FR

SocketPoll
`type` - removed
`upvotes` - removed
`answers` - will be `{ googleID : PollChoice }` instead of `PollChoice[]`

ClientPoll
`type` - removed
`upvotes` - removed

Socket
`user/poll/fr/live` - event removed
`server/poll/upvote` - event removed

Issue #346, opened 2020-02-01
Remove remnants of Poll Filtering/Free Response


jackthmp:
Remove stuff like `isFilterActivated`

PR #347, opened 2020-02-01
Removed QuestionType field in Poll models

jackthmp:
## Overview

QuestionType was an enum used when we had Free Response in the app. Since we now only use Multiple Choice, it's no longer needed and not represented in the upcoming API spec changes (#335). 

## Changes Made

Removed `QuestionType` and any references.

## Test Coverage

- Tested on Local Server with Alicia's `alicia/remove-fr` branch; all seems to be in the clear.

## Next Steps (delete if not applicable)

- Won't merge in until after the API spec changes are updated on the Dev/Prod server.

- #335


Pollo/Supporting/Constants.swift
```diff
@@ -49,8 +49,6 @@ struct Routes {
     static let userDelete = "user/poll/delete"
     static let userDeleteLive = "user/poll/delete/live"
     static let userEnd = "user/poll/end"
-    static let userFreeResponseFilter = "user/poll/fr/filter"
-    static let userFreeResponseLive = "user/poll/fr/live"
```
jackthmp: just random free response stuff leftover, will tackle more in #346

Pollo/Views/Cards/PollMiscellaneousCell.swift
```diff
@@ -100,10 +100,7 @@ class PollMiscellaneousCell: UICollectionViewCell {
                 descriptionLabel.text = "Shared with group"
             }
             var unit: String
-            switch miscellaneousModel.questionType! {
-            case .multipleChoice:
-                unit = miscellaneousModel.totalVotes == 1 ? voteString : "\(voteString)s"
-            }
+            unit = miscellaneousModel.totalVotes == 1 ? voteString : "\(voteString)s"
```
jackthmp: i think this is some leftover upvoting stuff?? will look into and remove

Pollo/Models/ListDiffableModels/Poll.swift
```diff
@@ -108,19 +88,15 @@ class Poll: Codable {
     func getSelected() -> Any? {
         if userAnswers.isEmpty { return nil }
         guard let googleID = User.currentUser?.id, let answers = userAnswers[googleID], !answers.isEmpty else { return nil }
-        switch type {
-        case .multipleChoice:
-            guard let answer = answers[0].letter else { return nil }
-            return answer
-        }
+        guard let answer = answers[0].letter else { return nil }
```
kevinchan159: we can probably chain these two `guard let` and i know we check for `!answers.isEmpty` but still might be safer to do `let answer = answers.first?.letter`

Pollo/ViewControllers/PollBuilderViewController.swift
```diff
@@ -264,53 +264,51 @@ class PollBuilderViewController: UIViewController {
...
-                        switch result {
-                        case .value:
-                            self.getDrafts()
-                        case .error(let error):
-                            print("error: ", error)
-                        }
+
+        let question = mcPollBuilder.questionText ?? ""
+        var options = mcPollBuilder.getOptions()
+       if options.isEmpty {
```
kevinchan159: is indentation off here?

jackthmp: Fixed!

Pollo/Models/ListDiffableModels/Poll.swift
jackthmp: Yes

Pollo/ViewControllers/PollBuilderViewController.swift
```diff
@@ -264,53 +264,51 @@ class PollBuilderViewController: UIViewController {
     // MARK: - Actions
     @objc func saveAsDraft() {
         if !canDraft { return }
-        guard let type = questionType else {
+        if !didLoad {
             print("warning! cannot save as draft before viewDidLoad")
```
chalo2000: curious about what issue this code was fixing

Issue #336, opened 2019-11-19
Student Answer Not Persistent for Running Poll


jackthmp:
After voting in an active poll, when you leave and come back to the poll the previously-selected answer is not present in the CardCell.

jackthmp:
Issue still occurs when exiting only to PollsDateViewController and returning to poll. Does not occur when leave class completely and re-enter.

jackthmp:
Also happens when you vote on an open poll and scroll over to previous questions and the cell is reused

PR #349, opened 2020-02-03
[MC Persistence Bug Fixed]

chalo2000:
## Overview

Fixed the bug where multiple choice answers would be displayed when entering from the PollViews -> PollsDateView -> CardController, but would disappear whenever going from the CardController -> PollsDateView -> CardController

## Changes Made

Originally userAnswers was being updated using `userAnswers[googleID]?[0] = PollChoice(text: mcChoice)` but this is incorrect because `text` should be the choice name, not the MC letter. The issue was that letter was nil and so the isSelected property of all buttons would not be selected. Adding another parameter to pass the choice name and properly instantiating the PollChoice fixed the issue.

In addition, `updatedSelected` wasn't working properly, possibly due to some remnants from FR. The buggy implementation accessed the first element of the user's `userAnswers`, which is redundant because there is ever only one answer for MC. The new implementation now stores the user's choice into `userAnswers` as a list.

## Test Coverage

Manual testing
1. going back to PollDatesView and to CardController [DONE]
2. having cells reused [DONE]
3. new poll question ~~[WIP]~~ [DONE] thx to @matthewcoufal 

## Next Steps

Fix other bugs after completing this one

## Related PRs or Issues

#336 

## Screenshots

  
  pollo_demo

^ outdated, answer now persists when making new poll question

### DIFF
--- a/Pollo/Models/ListDiffableModels/Poll.swift
+++ b/Pollo/Models/ListDiffableModels/Poll.swift
@@ -149,7 +149,7 @@ class Poll: Codable {
 
     func updateSelected(mcChoice: String, choice: String) {
         guard let googleID = User.currentUser?.id else { return }
-        userAnswers[googleID]?[0] = PollChoice(letter: mcChoice, text: choice)
+        userAnswers[googleID] = [PollChoice(letter: mcChoice, text: choice)]
     }
 
 }
--- a/Pollo/Models/ListDiffableModels/Poll.swift
+++ b/Pollo/Models/ListDiffableModels/Poll.swift
@@ -147,9 +147,9 @@ class Poll: Codable {
         return letter == mcChoice
     }
 
-    func updateSelected(mcChoice: String) {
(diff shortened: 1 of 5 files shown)


Pollo/Models/ListDiffableModels/Poll.swift
```diff
@@ -147,9 +147,9 @@ class Poll: Codable {
         return letter == mcChoice
     }
 
-    func updateSelected(mcChoice: String) {
+    func updateSelected(mcChoice: String, choice: String) {
         guard let googleID = User.currentUser?.id else { return }
-        userAnswers[googleID]?[0] = PollChoice(text: mcChoice)
+        userAnswers[googleID]?[0] = PollChoice(letter: mcChoice, text: choice)
```
jackthmp: ugh the whole letter vs. text thing is so annoying. hopefully soon backend will update and we'll only use indices for multiple choice options. but in the meantime this is a fine fix

Issue #344, opened 2020-01-29
Re-add bars for professor-side poll results


jackthmp:
Due to user feedback, we want to show to the professor dynamic bars that are proportional to the number of student responses for that choice. This is already enabled for student-side polling once results are shared, but all professors see are percentages.

PR #356, opened 2020-02-10
[Re-add bars for professor-side poll results]

chalo2000:
## Overview

The professor bars weren't proportional like the student ones, so I added that back.

**Professor side is now also redesigned**

## Changes Made

I moved some things in and out of a switch statement to give the professor the students' proportional bars. There are, **however**, other issues to resolve before integrating this into the app.

Image 1)
Just re-enabling bars doesn't look very nice; the corner radius on the highlightView bar clashes with the rectangular outer bar of the containerView. The background colors of the behind the bars also make the correct answer bar blend in too much and the light gray bars clash against the darker background.

Image 2)
This is after I fixed the issues from above. It looks nicer now.

Image 3)
The issue with this approach is that without the green background, I cannot tell what was originally the right answer unless at least one person answers it. In this case, it is obvious that the first option is the correct answer since the second option is not green, but what would we do in cases of more than 2 choices to choose from and at least 2 choices with no answers at all?

I feel we need to get some designs to fix this.

**Lots of refactoring in MCResultCell to make code more succinct and clearer.**
## Test Coverage

Manual testing (see images)

## Next Steps

After figuring out the proper approach to resolve the new design issue, resolve more Github issues.

**A couple of bugs discovered when implementing the new designs, especially for making draft questions (like #363). Will be putting up new issues to record these.**

## Related PRs or Issues

#344 

## Screenshots

<details>

<summary>Old screenshots</summary>

<details>

  <summary>1 Just re-enabling bars</summary>

Before mods

  
  
</details>

<details>
  <summary>2 Modifying the bars a little</summary>
After mods

</details>

<details>
  <summary>3 Issue with new modification</summary>
After mods issue

</details>
</details>

<details>
<summary>Student gets right answer</summary>

ezgif com-resize
</details>

<details>
<summary>Student gets wrong answer</summary>

ezgif com-resize (1)
</details>

<details>
<summary>Professor side with 2 devices changing their answers</summary>

ezgif com-resize (2)
</details>


Pollo/Views/Cards/MCResultCell.swift
```diff
@@ -169,48 +169,44 @@ class MCResultCell: UICollectionViewCell {
...
                 highlightView.backgroundColor = .lightGrey
                 highlightView.layer.borderColor = UIColor.coolGrey.cgColor
             }
+        } else {
+            if isSelected {
+                selectedDotView.backgroundColor = .coolGrey
+                selectedImageView.image = nil
+            }
+            highlightView.backgroundColor = .lightGrey
+            highlightView.layer.borderColor = UIColor.coolGrey.cgColor
```
YKo20010: Is there some way to reduce the nested if/else?
No idea if this works, but something like this:
```
highlightView.backgroundColor = .lightGrey
highlightView.layer.borderColor = UIColor.coolGrey.cgColor
selectedDotView.backgroundColor = isSelected ? .coolGrey : .clear
selectedImageView.image = nil
if let correctAnswer = correctAnswer, !correctAnswer.isEmpty {
              let selectedCorrectAnswer = answer == correctAnswer
              if isSelected {
                      selectedDotView.backgroundColor = .clear
                      selectedImageView.image = UIImage(named: selectedCorrectAnswer  ? correctImageName : incorrectImageName)
              }
              if selectedCorrectAnswer {
                  showCorrectAnswer = true
                  highlightView.backgroundColor = .lightGreen
                  highlightView.layer.borderColor = UIColor.polloGreen.cgColor
              }
          }
```

chalo2000: lol I didn't even realize how redundant the code was before

Pollo/Views/Cards/MCResultCell.swift
```diff
@@ -150,75 +137,107 @@ class MCResultCell: UICollectionViewCell {
...
-                highlightView.layer.borderColor = UIColor.coolGrey.cgColor
+        
+        numSelectedLabel.isHidden = userRole == .member
+        numSelectedLabel.text = "\(resultModel.numSelected)"
+        
+        optionLabel.text = resultModel.option
+        optionLabel.textColor = userRole == .admin ? .black : .mediumGrey
+        
+        percentSelected = resultModel.percentSelected
+        percentSelectedLabel.isHidden = userRole == .member
```
jackthmp: Can you sync up with design and check if we want to include the percentages/counts for the student-side?

Pollo/Views/Cards/MCResultCell.swift
```diff
@@ -150,75 +137,107 @@ class MCResultCell: UICollectionViewCell {
...
 
         highlightView.snp.makeConstraints { make in
             make.leading.top.bottom.equalToSuperview()
             make.width.equalToSuperview().multipliedBy(percentSelected)
         }
+        
+        numSelectedLabel.snp.makeConstraints { make in
+            make.top.equalToSuperview().inset(numSelectedLabelTopPadding)
+            make.trailing.equalToSuperview().inset(numSelectedLabelTrailingPadding)
+            make.width.equalTo(numSelectedLabelWidth)
```
jackthmp: do we need to set the width here and for `percentSelectedLabel`? can you just set leading = containerView and trailing = superView with an offset on either side?
just because we set the containerView's trailing to be inset from the superview with `adminTrailingPadding` so there should be enough space for the `numSelectedLabel` and `percentSelectedLabel`, it just seems redundant to have all these constants for the widths

Pollo/Views/Cards/MCResultCell.swift
```diff
@@ -163,13 +159,13 @@ class MCResultCell: UICollectionViewCell {
         
         numSelectedLabel.snp.makeConstraints { make in
             make.top.equalToSuperview().inset(numSelectedLabelTopPadding)
-            make.trailing.equalToSuperview().inset(numSelectedLabelTrailingPadding)
-            make.width.equalTo(numSelectedLabelWidth)
+            make.leading.equalTo(containerView.snp.trailing)
+            make.trailing.equalToSuperview()
```
jackthmp: Shouldn't there be some degree of margin between the label and the leading and trailing views? Like at least 4 pixels on either side just to take care of the case where there were 4 digits and it was squashed against the containerView?

Pollo/Views/Cards/MCResultCell.swift
```diff
@@ -163,13 +159,13 @@ class MCResultCell: UICollectionViewCell {
...
         }
         percentSelectedLabel.snp.makeConstraints { make in
             make.top.equalToSuperview().inset(percentSelectedLabelTopPadding)
-            make.trailing.equalToSuperview().inset(percentSelectedLabelTrailingPadding)
-            make.width.equalTo(percentSelectedLabelWidth)
+            make.leading.equalTo(containerView.snp.trailing)
+            make.trailing.equalToSuperview()
```
jackthmp: Same as above

Issue #355, opened 2020-02-09
Banner gets stuck when popping to PollsDateViewController


jackthmp:
ugh

PR #362, opened 2020-02-16
Fixed banner persistence issue

jackthmp:
(This is the right branch!)

## Overview

Fixed issue in which banners would appear and get stuck on `PollsViewController` when presented after a delay/connection after `PollsDateViewController` was popped. 

## Changes Made

Conformed banners to a delegate, added a check if the `currentBanner` is nil, if so then hide the new banner.

## Test Coverage

Manual testing, pop between VCs when banners are appearing.

#355

### DIFF
--- a/Pollo/Supporting/BannerController.swift
+++ b/Pollo/Supporting/BannerController.swift
@@ -35,6 +35,7 @@ public class BannerController {
 }
 
 extension BannerController: NotificationBannerDelegate {
+
     public func notificationBannerWillAppear(_ banner: BaseNotificationBanner) {
         banner.isHidden = self.currentBanner == nil
     }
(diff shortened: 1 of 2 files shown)


Pollo/Supporting/BannerController.swift
```diff
@@ -32,3 +33,21 @@ public class BannerController {
...
+    public func notificationBannerDidAppear(_ banner: BaseNotificationBanner) {
+        if self.currentBanner == nil {
+            banner.dismiss()
+        }
+    }
+
+    public func notificationBannerWillDisappear(_ banner: BaseNotificationBanner) { }
+
+    public func notificationBannerDidDisappear(_ banner: BaseNotificationBanner) { }
+
```
kevinchan159: remove newline

Pollo/Supporting/BannerController.swift
```diff
@@ -32,3 +33,21 @@ public class BannerController {
         }
     }
 }
+
+extension BannerController: NotificationBannerDelegate {
```
kevinchan159: add newline after here

Issue #348, opened 2020-02-02
Remove Attendance Exporting in Group Controls


jackthmp:
We decided to just contain this feature on web as there isn't enough need for it on mobile.

jackthmp:
Tbh can probably combine with #346 and just refactor the whole group controls page, pending designs

PR #374, opened 2020-02-26
Disabled group controls page

jackthmp:
## Overview

For now, we are disabling the group controls/settings page as we are not offering exporting from mobile. 

## Changes Made

- Hid arrow on admin navigation bar
- Disabled nav bar tap action to do nothing

## Test Coverage

Manual testing

## Next Steps

- `GroupControlsViewController` is preserved but currently is just dead code. When we add back location restrictions and other settings, we will reuse this code. 
- We will reenable the arrow and tap action when we add these settings back in.

## Related PRs or Issues

#348 

## Screenshots

<details>

  <summary>No more arrow in navigation bar on admin side</summary>

  Simulator Screen Shot - iPhone 11 Pro - 2020-02-26 at 15 26 49
  

</details>

### DIFF
--- a/Pollo/ViewControllers/PollsDateViewController+Extension.swift
+++ b/Pollo/ViewControllers/PollsDateViewController+Extension.swift
@@ -120,7 +120,9 @@ extension PollsDateViewController: PollBuilderViewControllerDelegate {
 extension PollsDateViewController: NavigationTitleViewDelegate {
 
     func navigationTitleViewNavigationButtonTapped() {
-        guard userRole == .admin else { return }
+        guard userRole != .admin && userRole != .member else { return }
+
+        // This code will be executed when we reenable the settings page
         let pollsDateAttendanceArray = pollsDateArray.map { PollsDateAttendanceModel(model: $0, isSelected: false) }
         getMembers(with: session?.id ?? "").observe { [weak self] result in
             guard let `self` = self else { return }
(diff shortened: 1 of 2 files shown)


Pollo/ViewControllers/PollsDateViewController+Extension.swift
```diff
@@ -120,22 +120,8 @@ extension PollsDateViewController: PollBuilderViewControllerDelegate {
 extension PollsDateViewController: NavigationTitleViewDelegate {
 
     func navigationTitleViewNavigationButtonTapped() {
-        guard userRole == .admin else { return }
```
chalo2000: would be good to remember we have this stored somewhere so we don't have to rewrite it

jackthmp:
> hahaha minor, but this branch should be `jackt/disable-settings`

omg i always use underscores idk why

Pollo/ViewControllers/PollsDateViewController+Extension.swift
jackthmp: i don't think we ever will rewrite it because it's all like past session fetching which we won't even have to do for the future settings page

Pollo/ViewControllers/PollsDateViewController+Extension.swift
```diff
@@ -120,22 +120,8 @@ extension PollsDateViewController: PollBuilderViewControllerDelegate {
...
-                    self.navigationController?.pushViewController(groupControlsVC, animated: true)
-                case .error(let error):
-                    print(error)
-                    let alertController = self.createAlert(title: "Error", message: "Failed to load data. Try again!")
-                    self.present(alertController, animated: true, completion: nil)
-                }
-            }
-        }
+        // TODO: Push the group controls page once we re-enable
+        return
```
kevinchan159: u can delete the return

PR #387, opened 2020-04-11
Updated API Spec

jackthmp:
## Overview

Updated API to reflect Backend changes listed here:
https://paper.dropbox.com/doc/Pollo-Backend-To-Do--Ax7uNitdezy647H7IaY5Gqk9Ag-LSAArQ6I0EzNN0LYQif7B

Previous PR #347 was approved for removing QuestionType.

## Changes Made
- Removed QuestionType (#347)
- Deprecated PollChoice, now represented by just an index of the MC option
- MC options are now represented by the index of the choice instead of the text or letter as a String.
- Modified PollResult, `letter` is no longer optional, now is an `int` represents the index of the option
- `correctAnswer` is now an int and defaults to -1 on the Backend, reflected in our models

## Test Coverage

- Ran server locally and everything worked.

## Next Steps (delete if not applicable)

- Will only merge in this PR once the API changes are reflected on Android and pushed to dev/prod

#335


Pollo/Extensions/Utils.swift
```diff
@@ -138,7 +130,8 @@ func formatResults(results: [String: JSON]) -> [String: PollResult] {
     results.forEach { (key, value) in
         let text = value[ParserKeys.textKey].stringValue
         let count = value[ParserKeys.countKey].intValue
-        pollResults[key] = PollResult(text: text, count: count)
+        let letter = value[ParserKeys.letterKey].intValue
```
chalo2000, line 133: instantiate constants in the same order as the passed arguments?

Pollo/Models/ListDiffableModels/Poll.swift
```diff
@@ -13,52 +13,24 @@ enum PollState: String, Codable {
...
-
 class PollResult: Codable, Equatable {
 
     static func == (lhs: PollResult, rhs: PollResult) -> Bool {
         return lhs.letter == rhs.letter && lhs.text == rhs.text && lhs.count == rhs.count
     }
 
     var count: Int?
-    var letter: String?
+    var letter: Int
```
chalo2000, line 23: probably would be a good idea to update this variable name so we don't intuitively deduce letter's type as a string

Pollo/Models/ListDiffableModels/Poll.swift
```diff
@@ -68,29 +40,27 @@ struct PollFilter: Codable {
 class Poll: Codable {
 
     var answerChoices: [PollResult]
-    var correctAnswer: String?  // only exists for multiple choice (format: 'A', 'B', ...)
+    var correctAnswer: Int
```
chalo2000, line 43: ditto

Pollo/Models/ListDiffableModels/Poll.swift
jackthmp: I have to keep it labeled as `letter` to match the API models on the backend side, otherwise i'd have to rename the keyvalue for Codable conformance

Pollo/Extensions/Utils.swift
jackthmp: Fixed

Pollo/Models/ListDiffableModels/Poll.swift
jackthmp: Fixed! backend renamed the field :)

Issue #388, opened 2020-04-12
Button when starting poll is higher than buttons of ended poll


chalo2000:
And also scrolling to other polls with begin to mess up the heights of the other buttons, probably for the same reason as #366 (dequeueing issue?)

chalo2000:
actually i made a new poll group and this doesn't seem like an issue - i've been reusing the same poll group for a while and perhaps some artifact is causing these issues

for example,
![Simulator Screen Shot - iPhone 8 - 2020-04-11 at 22 06 33](https://user-images.githubusercontent.com/21962778/79058838-d31f0c80-7c40-11ea-9413-e3102b2e79ca.png)

the gradient view is raised, but it's possible this has to do with the fact that the branch I'm working on lowered the button to stay consistent with Figma.

chalo2000:
not an coincidence, for some reason the constraints are acting real funky

Issue #383, opened 2020-03-12
Add emphasis to End Poll + Share Results button


chalo2000:
End Poll button should be red and Share Results should be green (see Renee's designs on Zeplin)

PR #389, opened 2020-04-19
Add button emphasis and fix responsive bugs

chalo2000:
## Overview

Added different colors to poll buttons to add emphasis to choices and fixed bugs not caught from before.

## Changes Made

- End Poll button is now red
- Button height is consistent after giving timer label a height anchor (for when the label has no text and defaults to zero height) 
https://stackoverflow.com/questions/45885086/why-uilabels-height-become-zero-if-doesnt-hold-any-text
- Removed questionCellHeight from constants (as it varies depending on text size)
- Fixed responsiveness to use appropriate questionCellHeight

## Test Coverage

Manual testing

## Related PRs or Issues

#383
#388 - may still have issues with OptionsGradientView if not related to button issue; if so, no consistent way to replicate found

## Screenshots

<details>
<summary>Button Emphasis</summary>

emphasis
</details>
<details>
<summary>Fixed responsiveness</summary>

fixed padding
</details>
<details>
<summary>Responsiveness with long questions</summary>

still works!
</details>

### DIFF
--- a/Pollo/Views/Cards/CardCell.swift
+++ b/Pollo/Views/Cards/CardCell.swift
@@ -273,7 +273,9 @@ extension CardCell: PollOptionsSectionControllerDelegate {
     }
     
     var maxCellHeight: CGFloat {
-        let cardCellTopHeight = LayoutConstants.hamburgerCardCellHeight + LayoutConstants.questionCellHeight + LayoutConstants.pollMiscellaneousCellHeight + LayoutConstants.separatorLineCardCellHeight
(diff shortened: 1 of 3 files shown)


Pollo/Views/Cards/CardCell.swift
```diff
@@ -113,6 +113,7 @@ class CardCell: UICollectionViewCell {
         timerLabel.snp.makeConstraints { make in
             make.centerX.equalToSuperview()
             make.bottom.equalToSuperview().inset(timerLabelBottomPadding)
+            make.height.equalTo(timerLabelFontSize)
```
jackthmp, line 116: The label height should equal 17.0 (as the area is on Zeplin), just to avoid any cropping of the text when you constrain to the exact font size.

Issue #371, opened 2020-02-24
Deleting last non-live question doesn't return to PollsDateView


chalo2000:
![ezgif com-resize (3)](https://user-images.githubusercontent.com/21962778/75128395-e296c600-5691-11ea-8dec-2ef9f78ac79c.gif)

Observe how the no polls message is also off-center

jackthmp:
I think this is intentional, although I'm not sure why it goes back if the question is deleted while live vs. ended. For example if you're in class you'd want to always be in CardController waiting for new questions even if they all were deleted, otherwise you'd have to tap back into the Date from PollsDate. Also, this case only looks weird when there was only 1 poll ever for this group, so both PollsDate and CardController are empty.
The offset is due to the collectionView having a left contentOffset for the regular cells and the empty cell is just added to the CV when empty. We can mess around and maybe shift the message over or something

PR #397, opened 2020-04-27
More Misc. Bug Fixes

jackthmp:
## Overview

- Fixed issue where banners were still being called when the presenting VC no longer exists
- Fixed strange banner race condition where banners were shown but not preserved as the `currentBanner`
- Fixed off-center empty states in CardController and prevented popping back to PollsDate when a live poll is deleted for consistency

## Test Coverage

Manual testing

#371

### DIFF
--- a/Pollo/Extensions/NotificationBanner+Shared.swift
+++ b/Pollo/Extensions/NotificationBanner+Shared.swift
@@ -13,7 +13,9 @@ import NotificationBannerSwift
 extension NotificationBanner {
 
     override public func isEqual(_ object: Any?) -> Bool {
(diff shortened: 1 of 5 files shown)


Pollo/Extensions/NotificationBanner+Shared.swift
```diff
@@ -12,6 +12,11 @@ import NotificationBannerSwift
 
 extension NotificationBanner {
 
+    override public func isEqual(_ object: Any?) -> Bool {
+        guard let banner = object as? NotificationBanner, let title1 = self.titleLabel?.text, let title2 = banner.titleLabel?.text else { return false }
```
kevinchan159, line 16: lets split across 3 lines:
```
guard let banner = object as? NotificationBanner, 
  let title1 = self.titleLabel?.text, 
  let title2 = banner.titleLabel?.text else { return false }
```

Issue #401, opened 2020-04-28
No Internet Screen trap for new users


chalo2000:
If I open the Google signin page and swipe the modal down, I'm stuck in the No Internet screen for good without reseting the app.

PR #403, opened 2020-04-28
Fix no internet edge case

chalo2000:
## Overview

Previously, any call to the Google login signin function would force the No Internet screen to appear if an error occurred. This is typically fine, except when new users accidentally cancel showing or dismiss the Google login modal, forcing them to be stuck in the No Internet screen unless they reset the app.

## Changes Made

Only show the No Internet screen if didSignInSilently, which is only set to true when the user has already authenticated in the past (aka, not new users). Otherwise, do nothing to keep new users on Pollo's sign in screen.

## Test Coverage

Manual testing

## Related PRs or Issues

#401

### DIFF
--- a/Pollo/AppDelegate.swift
+++ b/Pollo/AppDelegate.swift
@@ -192,7 +192,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, GIDSignInDelegate {
                 }
             }
             
-        } else {
+        } else if(didSignInSilently){
             self.pollsNavigationController.pushViewController(NoInternetViewController(), animated: false)
         }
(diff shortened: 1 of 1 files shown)


Pollo/AppDelegate.swift
```diff
@@ -192,7 +192,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, GIDSignInDelegate {
                 }
             }
             
-        } else {
+        } else if(didSignInSilently){
```
kevinchan159, line 195: `else if didSignInSilently {`

